PR #8, opened 2012-05-25
Added documentation about logging levels and extra data.

kof:
Here you can find logging levels

https://github.com/dcramer/sentry/blob/master/sentry/conf/defaults.py

I am using already both in production, it works for me.


benvinegar:
We've since added these notices to `/docs` – which are published to https://docs.getsentry.com.

PR #79, opened 2014-05-27
Add request feature detection

ruimarinho:
- Add several tests for each request property available on the HTTP
  interface.
- Add support for request abstractions such as Koa and Express.
- Improve error handling for malformed or incomplete request objects.
- Add support for `X-Forwarded-Port` for more robus HTTPS detection.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -1,5 +1,4 @@
 var cookie = require('cookie');
-var isPlainObject = require('lodash._shimisplainobject');
 var urlParser = require('url');
 var utils = require('./utils');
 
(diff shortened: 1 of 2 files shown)


wbyoung:
It looks like this reverted the changes made in #78. What's the desired way to set the user when running behind a proxy that sets `X-Forwarded-For` now? To write your own middleware like:

```js
/* @flow */

import Raven from 'raven';
import type {
  $Request as Request,
  $Response as Response,
  NextFunction,
  Middleware,
} from 'express';

export default function sentryContextMiddleware(): Middleware {
  return (req: Request, res: Response, next: NextFunction) => {
    const remoteIp =
      (req.headers['x-forwarded-for'] || '').split(',')[0] ||
      (req.connection && (req.connection: any).remoteAddress);

    const user: { [string]: mixed } = {
      ip_address: remoteIp, // eslint-disable-line camelcase
    };

    Raven.mergeContext({ user });

    next();
  };
}
```

Should that (or is that) documented somewhere?

PR #85, opened 2014-07-28
Replaced lsmod by package-list

ruimarinho:
When using tools like `pm2`, the method used by `lsmod` is not capable of detecting the packages of the child process. This means that sentry will always receive the list of modules of pm2 and not of your project's own dependencies.

Here's an example for a project using `foo` and `package-list`as dependencies, plus `lsmod` installed but not added to the project's `package.json`:

```
Output from `lsmod`:
{ pm2: '0.7.7', 'coffee-script': '1.7.0' }
------
Output from `package-list`
{ foo: '1.0.0',
  'package-list': '0.0.1',
  lsmod: '0.0.3 (extraneous)' }
```

This is up for discussion as the async nature of `getModules` changes the return values of some functions.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -47,8 +47,7 @@ _.get_ident = function getIdent(result) {
     return result.id;
 };
 
-_.process = function process(kwargs) {
-    kwargs['modules'] = utils.getModules();
+_.process = function process(kwargs, cb) {
     kwargs['server_name'] = kwargs['server_name'] || this.name;
     kwargs['extra'] = kwargs['extra'] || {};
     kwargs['extra']['node'] = process.version;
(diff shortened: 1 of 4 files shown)


kamilogorek:
Closing, as it's very outdated, over 3 years now. @ruimarinho please let me know if this is still something that you'd like to add and you could find time to verify it's still valid. Closing for now.

ruimarinho:
@kamilogorek the original issue is no longer valid. At first, I though the fix was done at the vendored file of `lsmod` that ships with raven, but using v0.0.3 (pre-PR) provides the same output. My guess is that the change in behaviour was introduced by pm2 (in the way it spawns child processes) or node itself.

For future reference, the current version of `lsmod` and its use inside `getModules()` is now able to correctly list the modules inside of the running script and not the launcher itself (in the original example, pm2 in either fork or cluster mode).

kamilogorek:
Thanks for getting back on this one. Cheers!

PR #93, opened 2014-09-28
Client takes additional request-context callback

visnup:
I needed to augment my exception data with user-based tags. I noticed sentry would automatically tag events with the number of affected distinct users if the client sends along user.email. This change provides a callback facility to augment the event with more data, based on the request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,19 +186,6 @@ app.listen(3000);
 
 __Note__: `raven.middleware.express` or `raven.middleware.connect` *must* be added to the middleware stack *before* any other error handling middlewares or there's a chance that the error will never get to Sentry.
 
-Both middlewares take an optional `parseRequest` function that will be called with the request and data being sent. It should return additional data and can be used to track user impact:
-
-```javascript
-app.use(raven.middleware.express('{{ SENTRY_DSN }}', function(req, data) {
-  data = data || {};
-
-  data.user = data.user || {};
-  data.user.email = req.person.email;
-
-  return data;
-}));
-```
-
 ## Coffeescript
 In order to use raven-node with coffee-script or another library which overwrites
 Error.prepareStackTrace you might run into the exception "Traceback does not support Error.prepareStackTrace being defined already."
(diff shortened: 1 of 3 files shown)


mattrobenolt:
So I don't like this approach being done on the middleware.

Does it make sense for this to be a configuration on the `Client` instead? Then have this function be called from within `parsers.parseRequest`.

That way this is a more general approach that can be used anywhere and piggy backs off of one entry point, `parsers.parseRequest`.

mattrobenolt:
So on second thought, looking at #97 and #103, we could just add some hooks to each step and let them be set globally on the `Client` config.

If this sounds reasonable, I'll happily accept a patch, or I can do it.

Sorry for the delay.

visnup:
I could see putting the config on the `Client`, but I think the reason we all gravitated to doing it on the middleware is because that's where connect-specific/request-based errors and raven come together.

I can move it to the `Client` config, but I don't quite follow with what you mean by "add some hooks to each step"..?

visnup:
@mattrobenolt I moved the callback function to the client configuration. calling it from the middleware still seems like the best place for it rather than down in `parsers.parseRequest`. is that what you meant?

jeffkole:
+1 on this.

night:
I could use this as well, and the last commit does fix the dependency on middleware. Thoughts @mattrobenolt

pwielgolaski:
:+1:, IMHO it is cleaner than using `client.setUserContext`

zwily:
What's the current thinking on this request? I'd really love to stop using my own fork for this functionality.

benvinegar:
@zwily – what are you trying to add, data-wise?

zwily:
@benvinegar - User and request data (user_id, request_id, etc). How are people doing this currently in express-land with Raven? Creating a new client for every request?

benvinegar:
User data can be set using `setUserContext`.

I think we'll merge in this patch, but I'll probably rename `requestCallback` to `dataCallback` – basically matching the API from raven-js (browser JavaScript).

zwily:
`setUserContext` is less useful when you're using a single client, and handling more than one request at once, right? Or am I missing something? Does `setUserContext` do some magic with domains under the covers or something?

benvinegar:
Okay, I see what you're saying. Most of this API is based off of raven-python, which actually asks you to set a user, then unset it after you've processed the response. But that's probably tricky to manage / error prone in Express world.

zwily:
Cool. 

FWIW, I also think this makes more sense on the middleware. The callback is going to supply the request object as a parameter (so you can get the current user from it), and that is specific to the middleware. I imagine that if I using 2 middlewares for some reason, I'd want to supply different functions since they will get different payloads.

But that's a very minor point. I won't complain if it's merged as-is. :)

LewisJEllis:
This was partially resolved a while back with the context overhaul, and the remaining missing use case for this should be handled by the new req parsing in #310.

PR #98, opened 2014-11-15
Disabled clients should trigger callbacks

mojodna:
Without this, boilerplate like...

``` javascript
raven.patchGlobal(function(success, err) {
  console.error(err.stack);
  process.exit(1);
});
```

...won't fire when the underlying client is disabled (i.e. no `SENTRY_DSN`).  However, the `uncaughtException` handler will still be registered, effectively swallowing errors.

This patch causes the client to pretend to have logged the request, allowing `patchGlobal` to trigger custom callbacks as would happen if it were enabled.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -67,7 +67,12 @@ _.process = function process(kwargs) {
     }
 
     // this will happen asynchronously. We don't care about it's response.
-    this._enabled && this.send(kwargs, ident);
+    if (this._enabled) {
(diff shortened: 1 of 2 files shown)


mojodna:
Bump.

LewisJEllis:
Similar to #138 #283 #301, resolved in #308, refer to [this comment](https://github.com/getsentry/raven-node/issues/138#issuecomment-299325889) for details.

PR #105, opened 2014-12-31
update to use /api/{projectId}/store

Raynos:
The default recommended rate limiting on the sentry
    documentation rate limits by `projectId`.

It's a better default to talk to a sentry server using the
    `/api/{projectId}/store` API.

This way multiple services talking to the same sentry
    server will not all get rate limited under the same
    projectId (which is presumably the empty string).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "0.6"
   - "0.8"
   - "0.10"
   - "0.11"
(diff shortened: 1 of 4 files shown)


dcramer:
That was a surprising pain in the ass to pull in :)

dcramer:
ref b725bd546b881a87a7b599f4b54a480b2234c36a

PR #106, opened 2015-01-04
Added support for RELEASE tag

SPARTAN563:
Included in Protocol v6 as defined in [the protocol docs](http://sentry.readthedocs.org/en/7.0.0/developer/client/index.html). This implementation pulls its information from the constructor's `options` or `process.env.RELEASE` or `process.env.VERSION` (in that order).
This makes the client fully compatible with Sentry 7+

I've included some basic tests as well as an addition to the README which provides a quick introduction regarding the use of the feature.

**TODO**
- Bump `package.json` version
- Release into the wild

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,10 @@ Optionally set the name for the client to use. [What is name?](http://raven.read
 ### SENTRY_SITE
 Optionally set the site for the client to use. [What is site?](http://raven.readthedocs.org/en/latest/config/index.html#site)
 
+### RELEASE
+Optionally set the application release version for the client to use, this is usually a Git SHA hash. Raven will also check to see whether
+the VERSION environment variable is set if it cannot find RELEASE.
+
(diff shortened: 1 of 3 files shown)


mattrobenolt:
I think we can leave out the environment variables for this. Setting `options.release` should be your application when you initialize the DSN.

If we do want to use environment variables for it, it'd like be something like `SENTRY_RELEASE` instead to not potentially clash with some unwanted env variable. But I assume we'd want to standardize that across the other clients first.

SPARTAN563:
Okay, I've switched it out for just `SENTRY_RELEASE` (though I can remove that as well if necessary). My primary reason for including VERSION is that a number of service managers set the VERSION variable, which would simplify interoperability with them - something which wouldn't be difficult to do within the init stage anyway.

dcramer:
Thanks, and sorry about the 5 month delay :)

Issue #108, opened 2015-02-09
There is no way to provide a custom SSL certificat for HTTPS


BlackYoup:
Hi there,

I'm trying to use raven using an HTTPS DSN with a self signed certificate. So, I'm getting a SELF_SIGNED_CERT_IN_CHAIN error when I try to emit messages to sentry.

Is there a way to add ``ca`` field to requests options ? http://nodejs.org/api/https.html#https_https_request_options_callback

I can try to make a PR but not sure about how to correctly give the arguments to the ``transport.send()`` method

Thanks !

mattrobenolt:
Hmm, if you wanna take a stab, I'm ok with that, if not I'll look into this later in the week and see how I can do this.

mattrobenolt:
Fixed with #110

BlackYoup:
Thanks :)

PR #110, opened 2015-02-10
Use custom certificate for HTTPS

BlackYoup:
References #108 

``ca`` parameter will be given to HTTPS requests. If http, it will don't change anything.

I tested it on my setup but I wasn't able to write tests. If you have an idea of to implement shuch tests, tell me and I'll write them.


mattrobenolt:
:+1:

mattrobenolt:
Thanks for the patch. :)

PR #114, opened 2015-03-07
Fix documentation about logging level and stackFunction

gigablah:
AFAIK setting the logging level on the client constructor doesn't do anything XD
Fixed the readme to put the level parameter on the correct line.

Also, looks like `stackFunction` was removed in 2b6586430dfc9ff35070ad9197be07a9855840ea, so I've removed it from the readme.


dcramer:
I pushed a ton of doc fixes today. I realize this is months old (sorry!), but wanted to close it out.

PR #123, opened 2015-08-03
Add passing extra options to HTTP transport

### LINKED ISSUE #115: Ability to specify custom options for transport
I have copy-pasted whole HTTPSTransport to be able to pass `rejectUnauthorized: false` to `https.request` (sorry for coffeescript):

```coffee
class HttpsTransport extends raven.transports.Transport
  defaultPort: 443
  transport: https

  send: (client, message, headers, ident) ->
    options =
      hostname: client.dsn.host
      path: client.dsn.path + 'api/store/'
      headers: headers
      method: 'POST'
      port: client.dsn.port || this.defaultPort
      rejectUnauthorized: false

    req = @transport.request(options, (res) ->
      res.setEncoding('utf8')
      if res.statusCode >= 200 && res.statusCode < 300
        client.emit('logged', ident)
      else
        reason = res.headers['x-sentry-error']
        e = new Error("HTTP Error (#{res.statusCode}): #{reason}")
        e.response = res
        e.statusCode = res.statusCode
        e.reason = reason
        e.sendMessage = message
        e.requestHeaders = headers
        e.ident = ident
        client.emit('error', e)

      # force the socket to drain
      res.on('data', noop)
      res.on('end', noop)
    )
```

It's not good. It would be great if I can just pass `rejectUnauthorized: false` to client's or transport's constructor.

maxvipon:
**Passing extra HTTP transport options**
```javascript
var raven = require('raven');
var client = new raven.Client('{{ SENTRY_DSN }}', {transportOptions: {rejectUnauthorized: false}});
 ```

fix #115

### DIFF
--- a/lib/transports.js
+++ b/lib/transports.js
@@ -6,9 +6,10 @@ function Transport() {
 util.inherits(Transport, events.EventEmitter);
 
 var http = require('http');
-function HTTPTransport() {
+function HTTPTransport(options) {
     this.defaultPort = 80;
     this.transport = http;
+    this.options = options || {};
 }
 util.inherits(HTTPTransport, Transport);
 HTTPTransport.prototype.send = function(client, message, headers, ident) {
@@ -19,7 +20,13 @@ HTTPTransport.prototype.send = function(client, message, headers, ident) {
         method: 'POST',
         port: client.dsn.port || this.defaultPort,
         ca: client.ca
-    }, req = this.transport.request(options, function(res){
+    };
+    for (var key in this.options) {
+        if (this.options.hasOwnProperty(key)) {
+            options[key] = this.options[key];
+        }
+    }
+    var req = this.transport.request(options, function(res){
         res.setEncoding('utf8');
         if(res.statusCode >= 200 && res.statusCode < 300) {
             client.emit('logged', ident);
@@ -46,9 +53,10 @@ HTTPTransport.prototype.send = function(client, message, headers, ident) {
 };
 
 var https = require('https');
-function HTTPSTransport() {
+function HTTPSTransport(options) {
     this.defaultPort = 443;
     this.transport = https;
+    this.options = options || {};
 }
 util.inherits(HTTPSTransport, HTTPTransport);
 
@@ -74,3 +82,5 @@ module.exports.http = new HTTPTransport();
 module.exports.https = new HTTPSTransport();
 module.exports.udp = new UDPTransport();
 module.exports.Transport = Transport;
+module.exports.HTTPTransport = HTTPTransport;
+module.exports.HTTPSTransport = HTTPSTransport;
--- a/README.md
+++ b/README.md
@@ -45,13 +45,18 @@ You can specify a level in the second optional parameter. Default level is `erro
 
 ```javascript
 var raven = require('raven');
-
 var client = new raven.Client('{{ SENTRY_DSN }}', {level: 'warning'});
 
-client.captureMessage("Another message")
+client.captureMessage('Another message')
 ```
 
-**Adding extra info an event**
+**Passing extra HTTP transport options**
+```javascript
+var raven = require('raven');
+var client = new raven.Client('{{ SENTRY_DSN }}', {transport: new raven.transports.HTTPSTransport({rejectUnauthorized: false})});
+```
+
+**Adding extra info to an event**
 ```javascript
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by mattrobenolt** [comment]
I think I'd prefer to see these be options that can be passed to the constructor of the transport, so if you need to override, you can just initialize your own `Transport` class.

Something like:

```javascript
Client(..., {transport: new HTTPSTransport({rejectUnauthorized: true})})
```

What do you think?

maxvipon:
Hm… Sounds right. It will look like this:
```js
Client(..., { transport: new raven.transports.HTTPSTransport({rejectUnauthorized: false}) })
```

maxvipon:
Done. Unfortunately, I have no idea how to test this…

lib/transports.js
```diff
@@ -1,14 +1,16 @@
 var events = require('events');
 var util = require('util');
+var assign = require('object-assign');
```
mattrobenolt: Can we drop this dependency? Seems like a silly that can just be replaced by a simple for loop.

mattrobenolt:
Other than my comment about removing `object-assign`, this all seems good. :+1:

mattrobenolt:
Oh, and squash your commits plz.

lib/transports.js
maxvipon: It's a simple for loop inside)) https://github.com/sindresorhus/object-assign/blob/master/index.js#L24-L39

I can copy-paste lib code to `lib/utils.js` but are you sure this is better?

mattrobenolt: How is that different than just?

```javascript
for (var key in this.options) this.options.hasOwnProperty(key)) && (options[key] = this.options[key]);
```

Or maybe I'm missing something on what this does. But we're just simply merging the two objects.

mattrobenolt: Or to be more verbose:

```javascript
for (var key in this.options) {
  if (this.options.hasOwnProperty(key)) {
    options[key] = this.options[key];
  }
}
```

maxvipon: In my opinion, the use of such small modules seems more clearly, it's node way.
You worry about extra dependency?

mattrobenolt: > You worry about extra dependency?

Yes. I don't want to have to traverse into external code and debug an issue for something so trivial. For example, for me to even figure out what this code was doing, I had to look this up.

I first google'd "object-assign", and it took me to: https://www.npmjs.com/package/object-assign Which is correct, then I was like, "I still don't understand what's going on here." Which brought me to https://github.com/sindresorhus/object-assign, and it's described as the "ES6 Object.assign() polyfill". Which links me to http://www.2ality.com/2014/01/object-assign.html. Since this wasn't official documentation, I ended up googling Object.assign, which brought me to the MDN page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign which finally says in plain English: `The Object.assign() method is used to copy the values of all enumerable own properties from one or more source objects to a target object. It will return the target object.`

:)

So yes, I'd prefer a simple 1 line for loop over this hunt trying to figure out what `assign` means personally.

It also may be because I'm old and don't know anything about ES6 so it wasn't immediately obvious what it did, and how it's different from what I proposed. I still don't actually know how it's much different except for the fact that it can merge multiple objects and we don't need that functionality.

mattrobenolt: Also, I apologize if that comes off as harsh. I just itemized the process I went through personally before I could understand the code I was looking at, which I don't feel is a very valuable addition to the project given the negative tradeoffs. I think a for loop is pretty simple and doesn't require an external dependency. If we could use `Object.assign()` in node, I'd :+1: it.

maxvipon: It's ok, I understand)) No, we can't use `Object.assign` now.

maxvipon: Removed

maxvipon:
@mattrobenolt take a look once again please and I squash It

maxvipon:
Squashed

mattrobenolt:
Awesome! Thanks, I'll double check this later today.

maxvipon:
:wave:

mattrobenolt:
derp

mattrobenolt:
Looks good, and thanks for the patience and the understanding. :smile: :cake: :sparkles:

maxvipon:
Thank you too! And waiting for release.

PR #125, opened 2015-08-23
Fix for callback flow to fix #86

### LINKED ISSUE #86: captureX() methods should call their callbacks after data has been successfully logged
The methods `captureError`, `captureMessage` and `captureQuery` call their callbacks to early: https://github.com/getsentry/raven-node/blob/master/lib/client.js#L128

In my code I use error domains to capture errors:

```javascript
var d = domain.create();

domain.run(function(){
  // code that throws errors
});
```

However if I use the `captureException` method, the events do not arrive at getsentry:

```javascript
domain.on('error', function(){
  sentry.captureException(err, function(){
    process.exit(1);
  });
});
```

The `process.exit` call will terminate the process to early. I think the behaviour of these methods should be similar to `patchGlobal` which calls its callback when the client emits the `logged` event.

Am I right, or did I miss something?

shamasis:
This PR fixes issue with the callback flow of `client.capture*` functions where it now gets called *after* the transporter has finished sending the logs to server.

Case in point:

The following code *does not* work since process.exit() kills all ongoing transports.

```javascript
client.captureError(err, {}, function () {
    process.exit(1);
});
```

This PR fixes this.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -80,8 +80,8 @@ _.process = function process(kwargs, callback) {
 
     // this will happen asynchronously. We don't care about it's response.
     this._enabled ? this.send(kwargs, ident, function (err) {
-        callback && callback(err, ident);
-    }) : (callback && callback(null, ident));
+        callback && callback(ident);
+    }) : (callback && callback(ident));
 
     return ident;
 };
--- a/lib/client.js
+++ b/lib/client.js
@@ -119,9 +119,8 @@ _.captureMessage = function captureMessage(message, kwargs, cb) {
     } else {
         kwargs = kwargs || {};
     }
-    var result = this.process(parsers.parseText(message, kwargs));
-    cb && cb(result);
-    return result;
+
+    return this.process(parsers.parseText(message, kwargs), cb);
 };
 
 _.captureError =
@@ -141,8 +140,7 @@ _.captureException = function captureError(err, kwargs, cb) {
         kwargs = kwargs || {};
     }
     parsers.parseError(err, kwargs, function(kw) {
-        var result = self.process(kw);
-        cb && cb(result);
+        self.process(kw, cb);
     });
 };
 
@@ -153,9 +151,8 @@ _.captureQuery = function captureQuery(query, engine, kwargs, cb) {
     } else {
         kwargs = kwargs || {};
     }
-    var result = this.process(parsers.parseQuery(query, engine, kwargs));
-    cb && cb(result);
-    return result;
+
+    return this.process(parsers.parseQuery(query, engine, kwargs), cb);
 };
 
 _.patchGlobal = function patchGlobal(cb) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -53,7 +53,7 @@ _.get_ident = function getIdent(result) {
     return result.id;
 };
 
-_.process = function process(kwargs) {
+_.process = function process(kwargs, callback) {
     kwargs['modules'] = utils.getModules();
     kwargs['server_name'] = kwargs['server_name'] || this.name;
     kwargs['extra'] = kwargs['extra'] || {};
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Hey @shamasis, this is actually an intentional decision and documented on how to handle it with events: https://github.com/getsentry/raven-node#events

There's an explicit reason why it's being done the way it is.

Is there some reason that the events can't cover your use case?

shamasis:
I would appreciate to know the reason for this code flow. As far as I see, using the events would result in using a very convoluted code flow.

1. captureError
2. on callback of capture error set a flag that an error has been captured
3. wait for logged or error to happen once on client
4. on those event of previous point, exit process.

It would be super tedious to carry around the `client` across all modules and our internal middlewares.

mattrobenolt:
We needed to return back an event id. This is similar to how the client works in other languages. We can return back an event id before the error is actually transmitted. The event id is calculated by the client.

Unfortunately, in node, the calculation of this event id requires reading files from disk, which need to be done in an asynchronous manner, which means, we need to leverage a callback instead of simply returning.

Ideally, the API would look like:

```
var eventId = client.captureError(...);
```

But that doesn't work. So I opted to make the event id be the callback. And the source of the error being actually transmitted or not, be dictated by the events being broadcast.

It's much less common for a developer to care when the event was actually transmitted successfully or not.

Typically, this is relevant for inside of an HTTP request and being used as a middleware, so exiting the process is not desirable. In the case where we do want to exit, say, in `patchGlobal`, the `patchGlobal` callback actually does trigger after the error has been transmitted, as documented here: https://github.com/getsentry/raven-node#catching-global-errors

You can also see how `patchGlobal` is implemented to produce the same thing on your end using normal `captureError`, etc: https://github.com/getsentry/raven-node/blob/master/lib/client.js#L163-L204

shamasis:
Appreciate the explanation. In other words, the `eventId` first approach is to better suit a more dominant use-case of raven. Understandable. This makes it do a lot of rewrites, like I mentioned. (Our use case is not similar to patchGlobal and is very specific to our code.)

Why did you not choose to resolve two async (transport + file read) and execute the callback with both? And does the code changes suggested by me break any future plans (I will not ask "what") :smile: ?

mattrobenolt:
>  In other words, the eventId first approach is to better suit a more dominant use-case of raven.

Correct.

> Why did you not choose to resolve two async (transport + file read) and execute the callback with both? 

Because we don't like to tie resolving an event id to the actual network transfer. Typically, this event id is utilized to render onto a page or something like that, and we don't want the sentry server to impact any flow. If the sentry server was unreachable, or network unstable and say, it took 5 seconds to send, I wouldn't want the client to delay 5 seconds just to render a 500 error page.

This could be encapsulated in some sort of promise though, but that'd probably require a bit more work. :)

Does this all make sense?

shamasis:
Sure. Having the promises would be exciting.

Keep up the great work. :smile:

jishi:
I don't understand this quote:

`Unfortunately, in node, the calculation of this event id requires reading files from disk`

Looking at the source, event_id is generated from a uuid, then set as ident.id, and in getIdent function it just returns ident.id?

```
kwargs.event_id = uuid().replace(/-/g, '');
...
var ident = {
    'id': kwargs.event_id
  };
...
proto.getIdent =
  proto.get_ident = function getIdent(result) {
    return result.id;
  };
```

The only thing that seems to prevent a synchronous return of the event id is that parsing is done before event id generation, something that probably is switchable.

mattrobenolt:
Yeah, sorry, I misspoke. The problem is order of operations as you pointed out. We parse the stacktrace which requires reading files from disk first, before assigning the event id.

jishi:
But is that really necessary? The most important thing to return when capturing errors is as I've understood, it, the event ID. The other stuff seems kind of redundant, and mostly for internal use.

mattrobenolt:
Likely not. This code is also super old, so it'd require looking into since I made an explicit decision here long ago, but maybe these constraints aren't required anymore as things have changed over the years.

PR #131, opened 2015-11-04
Fix tests for node 4.0

benvinegar:
Basically used events API instead of messing around w/ internals ... seemed to solve it.

https://nodejs.org/api/events.html

Tested with Node 0.10.x, 0.12.x, 4.0.x.


mattrobenolt:
thx :heart: :tada:

PR #132, opened 2015-11-05
Prevent off-heap memory explosion when request is stacking.

vernonzheng:
If request is stacking (e.g sentry server is down),  new ClientRequest will be added to agent.requests[]  and each ClientRequest uses Buffer. As clientRequest not consumed, process memory usage will explosive growth even though V8 memory usage is low.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -37,6 +37,7 @@ var Client = function Client(dsn, options) {
         this.ca = options.ca || null;
     }
 
+	this.maxReqQueueCount = options.maxReqQueueCount || 200;
     // enabled if a dsn is set
     this._enabled = !!this.dsn;
 
--- a/lib/transports.js
+++ b/lib/transports.js
@@ -26,6 +26,16 @@ HTTPTransport.prototype.send = function(client, message, headers, ident) {
             options[key] = this.options[key];
         }
     }
+	// in case sentry server is down
+	var _agent = options.agent || this.transport.globalAgent;
+	if (!_agent) {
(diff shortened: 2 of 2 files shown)


mattrobenolt:
Oh, this looks good. :)

Would it be possible to add a test for this, please? I'll review the code better in the morning.

mattrobenolt:
@benvinegar thoughts on this? It seems reasonable to me, but not sure if there's a better pattern or some semaphore type thing we can just pull in instead?

I wasn't familiar with this internal `globalAgent` object, which is pretty neat. Not sure if we should leverage that like in this patch, or if we should keep track of it ourselves inside the transport. Or if it matters.

lib/client.js
```diff
@@ -37,6 +37,7 @@ var Client = function Client(dsn, options) {
         this.ca = options.ca || null;
     }
 
+	this.maxReqQueueCount = options.maxReqQueueCount || 200;
```
benvinegar: Indentation

benvinegar:
The implementation of the patch seems sound, but we need tests before accepting. I also think, stylistically, we need a few things:

* Fix mismatched indentation.
* No underscore prefixed var names.

benvinegar:
@MaxBittker – can you add a test and see if we can get this landed?

MaxBittker:
status is that it's applied to master locally and passes existing tests, i just need to figure out how to write a new test for this

kamilogorek:
Moved to https://github.com/getsentry/raven-node/pull/369

PR #133, opened 2015-11-05
Push optional attributtes to captured error

diequeiroz:
Usage:

```javascript
// The request handler be the first item
app.use(raven.middleware.express.requestHandler(sentryDSN));

// The error handler must be before any other error middleware
app.use(raven.middleware.express.errorHandler(sentryDSN, function(attributes, req, error){
	attributes['sentry.interfaces.User'] = req.user;
	attributes.tags = {'env': 'production', 'version': '0.1'}
}));
```

### DIFF
--- a/lib/middleware/connect.js
+++ b/lib/middleware/connect.js
@@ -8,7 +8,7 @@ var connectMiddleware = function(client) {
 
 // Error handler. This should be the last item listed in middleware, but
 // before any other error handlers.
-connectMiddleware.errorHandler = function(client) {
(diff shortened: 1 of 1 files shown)


mattrobenolt:
We need to implement proper context in raven-node. There's no concept of that yet.

I'd rather us do that than implement this stop-gap solution imo. Going to create a new ticket to address that.

diequeiroz:
Thanks

gillesdemey:
:+1: to extend the context using the Express middleware

Issue #83, opened 2014-07-24
Implement context methods


dcramer:
We need to add the various context mechanisms to Node:

- user_context
- tags_context
- extra_context
- http_context

See the PHP client for some implementation examples:

https://github.com/getsentry/raven-php#providing-request-context

https://github.com/getsentry/raven-php/blob/master/lib/Raven/Context.php

benvinegar:
Added via #141

PR #141, opened 2016-01-23
Implement global context + add context API methods

### LINKED ISSUE #134: Implement proper context
See: #133 

raven-node has no concept of context like the other clients.

We need to expose `userContext`, `tagsContext`, `extraContext`, etc. along with the ability to clear. Then this can work for the middleware to clear after each request finish.

benvinegar:
Fixes #134, refs #83

Adds 3 new public methods, whose signatures matches the equivalent methods raven-js:
* `setUserContext`
* `setExtraContext`
* `setTagsContext`

Additionally, `extra` and `tags` can be specified during client initialization (again, like raven-js):

```javascript
var client = new raven.Client('DSN', {
  extra: { foo: 'bar' },
  tags: { browser: 'Chrome 46.0' }
});
```

Note that `user` cannot be specified via `raven.Client` – for no particular reason, other than that's what raven-js does and I'm trying to be as consistent as possible with that library.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -7,9 +7,20 @@ var uuid = require('node-uuid');
 var transports = require('./transports');
 var node_util = require('util'); // node_util to avoid confusion with "utils"
 var events = require('events');
-var extend = require('extend');
 
 module.exports.version = require('../package.json').version;
+
+var extend = Object.assign || function(target) {
+  for (var i = 1; i < arguments.length; i++) {
+    var source = arguments[i];
+    for (var key in source) {
+      if (Object.prototype.hasOwnProperty.call(source, key)) {
+        target[key] = source[key];
+      }
+    }
+  }
+  return target;
+};
 
 var Client = function Client(dsn, options) {
   if (arguments.length === 0) {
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -41,33 +42,47 @@ var Client = function Client(dsn, options) {
...
+  }
+  if (options.extra) {
+    globalContext.extra = options.extra;
+  }
 
   this.on('error', function(e) {}); // noop
 };
 node_util.inherits(Client, events.EventEmitter);
-var _ = Client.prototype;
+var proto = Client.prototype;
```
mattrobenolt: You no like my `_`?

benvinegar: Because it's synonymous with lodash/underscore. And nobody cares about saving 4 characters.

benvinegar:
> Could we do the polyfill thing here and utilize Object.assign if it exists? Since newer node supports that.

Since we're running tests against

benvinegar:
Oops, misclick.

bcallaars:
:shipit:

bcallaars:
Can we put a release out for this? Would be awesome if we could make it `0.9.1`?

benvinegar:
@bcallaars – I just published [0.10.0](https://github.com/getsentry/raven-node/releases/tag/0.10.0) to npm.

bcallaars:
Brilliant, great work!

kylehotchkiss:
@benvinegar thanks again ben :) Excited to give this a try this week!

PR #142, opened 2016-01-26
Improve raven-node docs ...

benvinegar:
- Remove install.rst – the information was deprecated (e.g. supported node versions), and it was not linked to except in the sidebar
- Add missing `var raven = require('raven')` in "Configuring the Client"
- Move "Configuring the HTTP Transport" to the bottom of usage.rst (shouldn't be at top of page)
- Remove redundant configuration example from usage.rst
- Add `Raven.captureException` example to usage.rst


benvinegar:
Hmm, I shouldn't remove the middleware / integrations link. Need to fit that back into the conversation somewhere.

mattrobenolt:
As you mentioned, the middleware/integrations link should be there somewhere. Other than that, it's fine.

Issue #150, opened 2016-04-29
Add fetch_git_sha in raven-node


cancan101:
Add a function like [`fetch_git_sha` from raven-python](https://github.com/getsentry/raven-python/blob/ed91eef804eb1453ca625270fe1b67e2b687b89b/raven/versioning.py) to the raven-node library.

benvinegar:
Better yet, let's document how to use an existing package like [git-rev](https://www.npmjs.com/package/git-rev) to set the release value in the raven-node docs.

kamilogorek:
Addressed in https://github.com/getsentry/raven-node/pull/382

PR #151, opened 2016-04-29
raven.utils.fetchGitSha as requested (#150)

damongant:


mattrobenolt:
Oh neat, this looks good! Lemme give it a closer look and go from there. Thanks. :)

lib/utils.js
```diff
@@ -177,6 +177,30 @@ function parseStack(err, cb) {
...
+  // just let it return an err if the file does not exist
+  fs.readFile(headPath, 'utf8', function(err1, headFileData) {
+    if (err1) {
+      return cb(err1);
+    }
+    if (headFileData.indexOf('ref: ') === 0) {
+      var ref = headFileData.split(' ')[1].trim();
+      var refPath = path.resolve(basePath, '.git', ref);
+      fs.readFile(refPath, 'utf8', function(err2, refFileData) {
+        if (err1) {
```
mattrobenolt: Shouldn't this be `err2`?

And while we're at it, where are there two err objects anyways? We could just mask the outer err within this inner scope since we don't need to access the outer scope.

DamonGant: eslint will complain about it. And yeah, I thought I forcepushed those along with the singlequote, just to realize it's still stuck at entering my passphrase :)

Give me a sec and I'll get a commit to fix those.

lib/utils.js
```diff
@@ -177,6 +177,30 @@ function parseStack(err, cb) {
...
+          return cb(err2);
+        }
+        cb(null, refFileData.trim());
+      });
+    } else {
+      cb(new Error('HEAD could not be parsed for git repository'))
```
mattrobenolt: For consistency, a semicolon here.

DamonGant: Done, any reason eshint isn't set up to check for those?

mattrobenolt: I didn't even know we set up eslint here, so that'd be a start! :)

/cc @benvinegar

DamonGant:
I'm not sure why the newer versions of node fail ([shallow cloning in older git versions?](http://stackoverflow.com/a/21217267/1754827)), but it seems I'll need to use the deprecated `Buffer` API for node 0.x either way.

Let me know if you'd prefer to use some kind of mock repo to test this, or maybe you can make travis use a more recent version of git.

mattrobenolt:
Bleh, I'm not sure off the top of my head. I'll have to look into this a bit. They also seem to be failing in an unrelated test. I'll get back to this later today for sure, just unfortunately, I can't merge this with broken tests until I understand why it's broken. :(

DamonGant:
No rush, this needs some additional work to be useful anyway.

mattrobenolt:
> No rush, this needs some additional work to be useful anyway.

Biggest issue we have is reading packed refs. After a `git gc`, the ref doesn't exist anymore and is packed away inside `.git/packed-refs`.

We need to add this into `raven-python` too, but shouldn't be too hard.

mattrobenolt:
I just threw up a PR for raven-python to handle parsing the packed-refs file, if you wanted to add that here as well: https://github.com/getsentry/raven-python/pull/761

It was more straightforward than I originally assumed.

kamilogorek:
There are some tools that can be used to do [that](https://github.com/tblobaum/git-rev) (or read directly from git cli) and I don't think it's necessary to add it to the SDK.
Please let me know if it's still desired feature though.

Issue #152, opened 2016-05-07
lsmod issue / TypeError: Cannot convert undefined or null to object


arabold:
I've just created a pull request for lsmod that fixes an issue in raven if no modules are loaded. I would appreciate if raven could update its dependencies once that pull request is merged. Thank you.

https://github.com/defunctzombie/node-lsmod/pull/2

arabold:
defunctzombie has updated [lsmod](https://github.com/defunctzombie/node-lsmod/) to 1.0.0 which fixes this bug. I've created a minimal pull request to update `lsmod`.

mattrobenolt:
Fixed with https://github.com/getsentry/raven-node/pull/154

PR #154, opened 2016-05-09
Update lsmod to 1.0.0

arabold:
This is for issue #152 to handle a null exception if no node modules are loaded (required).


mattrobenolt:
Thanks. :)

PR #164, opened 2016-05-29
Removes unnecesary single quotes in keys

willmendesneto:
Removing unnecessary single quotes in keys for objects. It's unnecessary because the keys don't have any special characters, so we can use JSON notation for them.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/164)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,8 +9,8 @@ var lsmod = require('lsmod');
 var stacktrace = require('stack-trace');
 
 var protocolMap = {
-  'http': 80,
-  'https': 443
+  http: 80,
+  https: 443
 };
 
 module.exports.getAuthHeader = function getAuthHeader(timestamp, api_key, api_secret) {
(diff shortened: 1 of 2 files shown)


DamonGant:
Personally I'm a fan of consistent singlequotes everywhere, but "only if needed" can be done too. 

Either should be picked as the standard style.

lib/utils.js
```diff
@@ -134,7 +134,7 @@ function parseStack(err, cb) {
     var frame = {
         filename: line.getFileName() || '',
         lineno: line.getLineNumber(),
-        'function': getFunction(line),
+        function: getFunction(line),
```
benvinegar: I think it's safer to leave this (notice it's highlighted as a keyword here). Revert this one line and I'll merge.

Thanks!

willmendesneto:
@benvinegar Done!

benvinegar:
Thanks!

PR #165, opened 2016-05-31
Add request parsing info to koa docs

rmi7:
When using the koa docs to implement Sentry logging in a koa.js app. there will be no request info.

To get this info in Sentry Events the request (`req`) should be parsed. 

This needs to be done manually, as can be seen in the pull request.

### DIFF
--- a/docs/integrations/koa.rst
+++ b/docs/integrations/koa.rst
@@ -21,7 +21,7 @@ Koa
 
   app.listen(8080);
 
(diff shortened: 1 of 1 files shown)


benvinegar:
@LewisJEllis – not sure if I'd use this patch as-is, but would be nice if we could verify if a) request information does indeed not work, and b) if so, add a simple how-to to the docs.

rmi7:
It has been 8 months almost so if you had answered earlier I would've checked, I'm just gonna close it. I've moved on. Great one on the response time 👍

PR #166, opened 2016-06-02
Use correct SPDX license expression syntax for BSD license

Starefossen:
This PR updates the `license` property in `pacakge.json` with the currect SPDX license expression syntax for the BSD license for this project `BSD-2-Clause`:

```
npm info package.json raven@0.11.0 license should be a valid SPDX license expression
```


mattrobenolt:
Thanks! :cookie:

PR #169, opened 2016-06-14
Whitelist passed process.env variables

psergus:
It creates a security hole if you pass init variables to your application through environment. If any security tokens are used, raven passes them through, which is an undesired behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "debugging",
     "exceptions"
   ],
-  "version": "0.11.0",
+  "version": "0.11.0-env",
   "repository": "git://github.com/getsentry/raven-node.git",
   "author": "Matt Robenolt <matt@ydekproductions.com>",
   "license": "BSD-2-Clause",
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -11,6 +11,17 @@ module.exports.parseText = function parseText(message, kwargs) {
   return kwargs;
 };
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "debugging",
     "exceptions"
   ],
-  "version": "0.11.0",
+  "version": "0.11.0-env",
```
mattrobenolt: Before looking at anythign else, please don't bump the version number. :)

psergus: fair enough. I will update it. thx.

mattrobenolt:
I'd +1 for just removing this functionality entirely out of `raven-node`. I just checked, and our python implementation doesn't capture environment variables at all by default.

psergus:
https://github.com/getsentry/raven-python/blob/ed91eef804eb1453ca625270fe1b67e2b687b89b/raven/utils/wsgi.py#L27

looks like python whitelists as well

psergus:
anyway, I will reopen PR from different branch. Will close this one. Sorry about that.

mattrobenolt:
Those are different. In python, specifically for web servers, these are related to the HTTP implementation, specifically a part of the WSGI spec. These aren't environment variables on the machine, or environment variables global for the process as in `process.env` here.

psergus:
I see your point. I guess I will rip out process.env from here and create a new PR for this. For now, we just added this whitelist functionality to make sure we do not pass anything secured.

PR #170, opened 2016-06-15
Whitelist passed process.env variables

psergus:
It creates a security hole if you pass init variables to your application through environment. If any security tokens are used, raven passes them through, which is an undesired behavior.


psergus:
@mattrobenolt here you go, process.env variables. Let me know if that's what you meant.

benvinegar:
Closed in favor of #181

PR #173, opened 2016-06-22
Add support for 'code' field to skip client errors

vturcotte:
We use 'code' as name for status codes in our API.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/173)

### DIFF
--- a/lib/middleware/connect.js
+++ b/lib/middleware/connect.js
@@ -13,7 +13,7 @@ var connectMiddleware = function(client) {
 connectMiddleware.errorHandler = function(client) {
   client = client instanceof raven.Client ? client : new raven.Client(client);
   return function(err, req, res, next) {
-    var status = err.status || err.statusCode || err.status_code || 500;
+    var status = err.status || err.code || err.statusCode || err.status_code || 500;
 
     // skip anything not marked as an internal server error
     if (status < 500) return next(err);


mattrobenolt:
Could we add this to the end of the list? Is there a reason this has to have such high priority?

mattrobenolt:
Also:

>  in our API

Is this a home rolled thing? Or is this a convention adopted by a framework? The ones that are there now all came from existing framework patterns.

vturcotte:
We can add it to the end of the list for sure.

vturcotte:
We use the jsend convention: https://labs.omniti.com/labs/jsend

mattrobenolt:
Cool, thanks. :)

I'll let @benvinegar give final approval and merge, but it sounds fine to me.

benvinegar:
Doing a little bit of research into this, it doesn't seem like JSend is broadly adopted, despite being around for years. Most libraries "implementing" the specification on GitHub don't have more than a dozen or so stars.

So, I'm hesitant to adopt this change, even if it is really minor. My feeling is that you should write your own `errorHandler` implementation in your code that looks at `err.code`. It's only ~20 lines.

LewisJEllis:
In addition to @benvinegar's low-adoption concerns, I noticed that the [jsend spec](https://labs.omniti.com/labs/jsend) and the [jsend npm module](https://github.com/Prestaul/jsend) make no indication that the `code` field is explicitly meant to be an http status code. They use example values like `42` and `123`, and the spec describes it as a "numeric code corresponding to the error." As such, I don't think we should interpret the `code` field like an http status code in deciding whether to capture an error.

PR #174, opened 2016-06-27
add cookie to request headers

dmcqueen:
Needed a way to send cookie headers in the http request.  There may be a better or existing way to solve this cleanly but this pull allows for some config like this...

```
var ravenClient = new raven.Client(SENTRY_DSN, {
    level: 'error',
    transport: new raven.transports.HTTPSTransport({ cookie: 'blah=blah; blah2=blah2' })
});
```

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/174)

### DIFF
--- a/lib/transports.js
+++ b/lib/transports.js
@@ -24,7 +24,9 @@ HTTPTransport.prototype.send = function(client, message, headers, ident) {
     ca: client.ca
   };
   for (var key in this.options) {
-    if (this.options.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


benvinegar:
I'd probably prefer `options.defaultHeaders` so you can specify more than just the `Cookie` header.

kamilogorek:
Agree with @benvinegar. @dmcqueen please let me know if you'd be willing to help with this feature. I'm more than happy to help.

Closing this PR for now.

PR #177, opened 2016-07-28
Copy `req.user` to user interface if present (fixes #160)

### LINKED ISSUE #160: How to add user data to the error object in express
I am adding user data to the req object in express as shown in this simplified example below. I am then using the dataCallback function to check to see if the user data has been added and it doesn't look like it has.

```
var raven = new Raven.Client( config.SENTRY_DSN, {
    dataCallback: function(data) {
        console.log(data.request.user.name); // undefined
        return data;}
});

app.get('/example', addUserData, function (req, res) {
    throw new Error('Broke!');
});

app.use(Raven.middleware.express.errorHandler(raven));

function addUserData(req, res, next){
    req.user = { name:'john', id:'1' }
    next();
}
```

However none of that user data seems to present on 'data' object that is passed into 'dataCallback' function. It seems more like the raw request before my custom middleware has added stuff to the object.

Am I missing obvious something here, is there another way of passing user data to Sentry in express? Do you have an example anywhere?

Thanks guys!

benvinegar:
cc @macqueen @mattrobenolt

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/177)

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -166,5 +166,26 @@ module.exports.parseRequest = function parseRequest(req, kwargs) {
   // expose http interface
   kwargs.request = http;
 
+  // user
+  //
+  // typically found on req.user according to Express and Passport
+
+  var user = {};
+  if (req.user && !kwargs.user) {
+    // shallow copy is okay because we are only modifying top-level
+    // object (req.user)
+    for (var key in req.user) {
+      if ({}.hasOwnProperty.call(req.user, key)) {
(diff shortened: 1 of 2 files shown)


macqueen:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/raven-node/177)*

PR #178, opened 2016-07-28
Don't mutate process.env when attaching REMOTE_ADDR (fixes #158)

### LINKED ISSUE #158: parseRequest modified global process.env
https://github.com/getsentry/raven-node/blob/master/lib/parsers.js#L164

Since http.env is only a reference to process.env, and not a copy of it, it will modify the global env object

benvinegar:
cc @macqueen @mattrobenolt

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var shallowCopy = require('shallow-copy');
 var cookie = require('cookie');
 var urlParser = require('url');
 var utils = require('./utils');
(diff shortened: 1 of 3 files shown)


macqueen:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/raven-node/178)*

benvinegar:
I just realized that a PR was opened to just straight-up remove collection of most ENV variables ... 
#169. So this will probably get re-written anyways.

PR #179, opened 2016-07-29
add in environment option

jbaxleyiii:
In order to support https://blog.getsentry.com/2016/07/22/environment-details.html

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/179)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -41,6 +41,7 @@ var Client = function Client(dsn, options) {
   this.root = options.root || process.cwd();
   this.transport = options.transport || transports[this.dsn.protocol];
   this.release = options.release || process.env.SENTRY_RELEASE || '';
+  this.environment = options.environment || process.env.SENTRY_ENVIRONMENT || '';
 
   this.loggerName = options.logger || '';
   this.dataCallback = options.dataCallback;
--- a/test/raven.client.js
+++ b/test/raven.client.js
@@ -127,6 +127,20 @@ describe('raven.Client', function() {
     delete process.env.SENTRY_RELEASE;
   });
 
+  it('should pull environment from options if present', function() {
+    var client = new raven.Client(dsn, {
+      environment: 'staging'
+    });
+    client.environment.should.eql('staging');
+  });
+
+  it('should pull SENTRY_ENVIRONMENT from environment', function() {
+    process.env.SENTRY_ENVIRONMENT = 'staging';
+    var client = new raven.Client(dsn);
+    client.environment.should.eql('staging');
+    delete process.env.SENTRY_ENVIRONMENT;
+  });
+
   describe('#getIdent()', function() {
     it('should match', function() {
       var result = {


benvinegar:
If you could add a cursory test for this I will merge pronto.

jbaxleyiii:
@benvinegar can do!

lib/client.js
```diff
@@ -41,6 +41,7 @@ var Client = function Client(dsn, options) {
   this.root = options.root || process.cwd();
   this.transport = options.transport || transports[this.dsn.protocol];
   this.release = options.release || process.env.SENTRY_RELEASE || '';
+  this.environment = options.environment || process.env.SENTRY_ENVIRONMENT || '';
```
mattrobenolt: Wouldn't we want to leverage `NODE_ENV` for this? Or is that not really a thing anymore?

benvinegar: I kind of agree, and was going to ping @dcramer whether SENTRY_ENVIRONMENT was a thing.

jbaxleyiii: @mattrobenolt @benvinegar I thought about that but saw from the blog:

> By default we automatically select the most production like environment

and I also saw [this](https://github.com/getsentry/raven-node/blob/aa6ac5082272d6dd1f41f8e0231fc754adb72974/History.md#050---482013) so I opted not too. I'm happy to add it in as a fallback!

fwiw, most people I know run node in 'production' for staging sites so I don't know how much value it would add

mattrobenolt: > Remove NODE_ENV entirely, fixes many issues since people have different opinions on wtf this means

lol, yeah, I remember this. Before, it was automatically enabling/disabling Raven as a whole if `NODE_ENV` wasn't production.

But yeah, you might be right. I don't think this is really used to mean an environment. I don't think someoen would say `NODE_ENV=staging`, etc.

jbaxleyiii:
@benvinegar tests added 👍

benvinegar:
Fantastic. Thanks!

fboutin-pmc:
Did anybody "functionally tested" the thing because I cannot see the `kwargs` being set in the `process` function?

I'm using `0.22.0` right now solely for this new feature and I'm trying to figure why it doesn't work for me...

fboutin-pmc:
Well, reading the Sentry documentation you can set the `environment` called `context` through the `tags` field.
> Source https://docs.getsentry.com/hosted/clientdev/attributes/

What I'm saying is that, setting the environment worked priori to `0.22.0` and you only need to pass some extra `options`. For example, when initializing the Client,

````
var raven = require('raven');
var client = new raven.Client(
    '<url>',
    {
        tags: {
            environment: 'development'
        }
    });
```
And it works, I get the `environment` tag in Sentry... Or may I miss something?

benvinegar:
>  I cannot see the kwargs being set in the process function?

Hmm, yeah, looking back ... not sure why I thought this was functionally correct. I'll follow-up with a fix.

> And it works, I get the environment tag in Sentry... Or may I miss something?

The environment tag is different from the explicit "environment" attribute. More here: https://docs.getsentry.com/hosted/clientdev/attributes/

benvinegar:
See #185

fboutin-pmc:
It works now with #185. 

But Sentry is quite weird with the `Environment` thing. I don't understand the difference between a `tag` and an `attribute`. Both are interpreted by Sentry but setting both produces some weird things, for example here on Slack,

screen shot 2016-08-05 at 09 01 22

dcramer:
If I had to guess I'd say you're sending "Environment" and "environment" and at this time tags are case sensitive.

dcramer:
Alternatively it might be that it's tagging it twice but most of our code doesn't support that. Either way you don't need to send the tag when you've sent the top level attribute.

fboutin-pmc:
Don't worry, I used the very same strings ;)

Ok then, now waiting for 0.23.0 I guess.

PR #180, opened 2016-07-29
Pass result of captureException-attempt to patchGlobal callback

JBLatenight:
Any reason it doesn't do this today?

Seems handy - could pass crash report ID to user in the error dialog or otherwise log, no?

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/180)


benvinegar:
At a glance this makes sense. Admittedly, I need to understand these callbacks a little better so this may take me some time before I can review properly.

JBLatenight:
👌👌

LewisJEllis:
We're now doing roughly this with the `sendErr` param of the `install` fatal error callback introduced in #308.

PR #185, opened 2016-08-04
Fix environment not attached from opts/kwargs

benvinegar:
cc @macqueen @fboutin-pmc

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/185)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -83,6 +83,7 @@ proto.process = function process(kwargs) {
     kwargs.extra.node = process.version;
   }
 
+  kwargs.environment = kwargs.environment || this.environment;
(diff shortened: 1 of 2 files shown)


benvinegar:
Besides tests, I've manually verified that `environment`, when configured, is properly send to a Sentry project.

fboutin-pmc:
We should not forget to test the default value we can set for the environment variable in Sentry project settings,

screen shot 2016-08-05 at 08 49 56

In other words, I don't know if we will need to do the same thing we do with the `release` field - i.e. avoid to assign it a value so we don't override the default one.

Issue #196, opened 2016-08-27
Explore having raven-node keep process running until error is transmitted


benvinegar:
Right now if you do a simple execute-once script to verify raven-node functions – you'll find that the process sometimes (always?) exits before the error payload is transmitted.

Even though their eventual goal is to use raven-node with an Express-based app, this is their first experience w/ Sentry and to them it may feel like Sentry just *doesn't work*.

Explore a solution that keeps the Node process running until raven-node is finished. It's probably ideal to introduce this as a configuration option (so that people can use it with build scripts, etc). And at least it will be documented.

mattrobenolt:
This is, I think, more relevant with Lambda now. But I could have sworn that Node won't shut down if there are outstanding async things happening until everything has finished.

benvinegar:
I've definitely experimented with this recently and the process exited first.

jakedahm:
I am experiencing this issue right now actually when trying to send logs from Lambda

LewisJEllis:
I played around with this a bit and I've not been able to reproduce this, even with the most minimum possible example:
```javascript
(new (require('raven')).Client('<dsn>')).captureException(new Error('errormsg'));
```

Following the code, we have synchronous calls from `client.captureException` all the way to a `zlib.deflate` with an async callback that calls `self.transport.send` which results in an `https.request` with an async callback which according to [docs](https://nodejs.org/docs/latest-v6.x/api/http.html#http_http_request_options_callback) is added as a listener for the `http.ClientResponse.response` event. Then node will keep things alive until that response comes back. The only scenario I can imagine where we would exit earlier is if we thought we sent the request, but we actually didn't, and so that listener isn't there. I don't have any ideas how that might actually occur.

Here's a bigger example with some more lifecycle logging:
```javascript
var raven = require('raven');
var client = new raven.Client('<dsn>', {
  dataCallback: function (data) {
    console.log('dataCallback fired');
    return data;
  }
});

process.on('beforeExit', function () {
  console.log('beforeExit fired');
});

client.on('logged', function () {
  console.log('logged error');
});

client.captureException(new Error('errormsg'));

console.log('run to completion');
```

I would be curious to see a reproducible example of the process exiting before the error is logged, or to see the output of running my above example on Lambda.

Regarding whether captureException should call its callback after the request is complete - is `client.on('logged')` not sufficient for that use?

LewisJEllis:
[...]

PR #200, opened 2016-08-31
Add Transport Callbacks (#196)

jakedahm:
In some cases the node process will exit before the request is sent. This is the case when using AWS Lambda. I am proposing that there is a callback added to the transports, which is called only after the request is complete.

``` js
client.captureException(new Error('oh no'), function() {
  // Async request is complete
  done(); // exit process
});
```

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/200)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -75,7 +75,7 @@ proto.getIdent =
     return result.id;
   };
 
-proto.process = function process(kwargs) {
+proto.process = function process(kwargs, cb) {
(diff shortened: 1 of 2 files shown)


benvinegar:
How does this work when an object is passed as the 2nd argument (for context, tags, etc)?

Also, this is easier to accept if there are tests.

LewisJEllis:
This should be addressed by #217.

PR #205, opened 2016-09-15
Update eslint to 2.13.1, make eslintrc more strict

LewisJEllis:
I made the eslintrc more strict and cleaned up a few minor things that that caught; it wasn't much so these code changes are pretty minor, but the PR is mostly to make sure the ESLint version situation makes sense:

We're using 2.13.1 here instead of latest 3.5.0 because ESLint 3.x dropped support for node < 4. With 3.5.0 travis would fail on 0.8-0.12, but we don't have a strong reason to need ESLint 3.x, so staying on 2.x was the easiest fix for now.

Unfortunately the Travis docs didn't have anything helpful to let us run a different test command for different node versions, so if we want to upgrade to ESLint 3.x, I think the best we could do is make the lint step figure out if it's on node < 4 and if so skip linting.

/cc @benvinegar

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/205)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,62 +3,86 @@
     "node": true,
     "mocha": true,
   },
+  "extends": "eslint:recommended",
   "rules": {
+    "array-bracket-spacing": [2, "never"],
     "block-scoped-var": 2,
-    "dot-notation": [
-      2,
-      {
-        "allowKeywords": true
-      }
-    ],
-    "eqeqeq": [
-      2,
-      "allow-null"
-    ],
+    "brace-style": [2, "1tbs"],
+    "camelcase": [1, "properties": "never"],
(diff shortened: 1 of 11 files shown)


benvinegar:
🚢 

> Unfortunately the Travis docs didn't have anything helpful to let us run a different test command for different node versions, so if we want to upgrade to ESLint 3.x, I think the best we could do is make the lint step figure out if it's on node < 4 and if so skip linting.

I agree, we don't need to run the linter on every Node version. But I don't think there's a strong impetus to upgrade ESLint at the moment either. When we need to, we'll figure out what Travis needs and make that change.

benvinegar:
🎉

PR #211, opened 2016-10-05
call next() inside onError example

chrisblossom:
I ran into an issue by removing the ``next`` function param from ``onError`` in the Express middleware documentation that was causing my app to crash when an error happened. I did this because Eslint was complaining about ``no-unused-vars``. 

I added ``next();`` to the end of the function and now everything is happy.

### DIFF
--- a/docs/integrations/connect.rst
+++ b/docs/integrations/connect.rst
@@ -15,6 +15,7 @@ Connect
       // and optionally displayed to the user for support.
       res.statusCode = 500;
       res.end(res.sentry+'\n');
+      next();
   }
 
   connect(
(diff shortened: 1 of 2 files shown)


LewisJEllis:
The `next` parameter is necessary due to [express error handling](https://expressjs.com/en/guide/error-handling.html); it inspects middleware functions and specifically [looks for four parameters](https://github.com/expressjs/express/blob/master/lib/router/layer.js#L62), and only then will treat them as error handlers.

In these particular examples the `next()` call doesn't do anything since there are no more error handlers to continue to, so I'm concerned that the presence of a `next()` call might be misleading. I wouldn't add it just because ESLint tells you to; there's really no need or reason for it.

My recommendation would be to configure [no-unused-vars](http://eslint.org/docs/rules/no-unused-vars) to not complain about unused arguments called `next`:
```javascript
"no-unused-vars": ["error", { "argsIgnorePattern": "next" }]
```
or you could do it as a one-off with something like:
```javascript
// eslint-disable-next-line no-unused vars
```

Refer also to expressjs/generator#78 which discusses this same issue.

Issue #213, opened 2016-10-11
Fix the API around capture* and handle callback correctly


mattrobenolt:
Due to legacy reasons, `capture*`'s callback was to get the eventId.

So the current API is:

```javascript
client.captureError(new Error(), function(result) {
  console.log(client.getIdent(result));
});
```

This entire pattern is basically deprecated and unnecessary now for Node.

Ideally, we'd have a pattern like this now:

```javascript
var eventId = client.captureError(new Error(), function(err) {
  console.log(err);
});
```

Where the callback now indicates when the event has been sent or failed to send. And there's no reason now that we can't return the event id immediately.

This should be done for a 1.0 imo since this is a major breaking API change, but for the better.

LewisJEllis:
Yea, I had realized we're capable of just immediately return the event id, but hadn't thought about the API/timing/related issues as much. Looking at them now I think this is the right move and I agree with the pattern you set out. It'll fit well with the other API changes happening in 1.0.0

Related: #200 #125 #196 #86

LewisJEllis:
Addressed by #217.

PR #217, opened 2016-10-20
Make capture* callbacks fire after sending, more API cleanup/singletonization

LewisJEllis:
Short version: addresses #213, moves the API to follow raven-js almost exactly, still maintains BC in almost all ways with old API

Details:
- captureException etc now fire their callbacks after sending to the server, as in #213 
- Moved everything to a Raven class instead of Client, following raven-js pattern/API
  - We now export a default Raven instance, leading to almost-identical usage to raven-js
  - Still have a Client wrapper for bc/multi-instance capabilities
- Does some other polishing/cleanups on the API changes
- Adds glue to maintain backwards compat with old API as much as possible
  - None of the changes to tests actually affect any existing public API usage patterns at all

The only public API usage affected by this should be the one described by @mattrobenolt in #213, and the realistic caveat to run into from it is that existing code looking at `eventId`s from `capture*` callbacks will now get `null` instead of a string.

/cc @benvinegar

### DIFF
--- a/test/raven.client.js
+++ b/test/raven.client.js
@@ -382,6 +382,36 @@ describe('raven.Client', function () {
       });
     });
 
+    it('should call the callback after sending', function (done) {
+      var firedCallback = false;
+      var sentResponse = false;
+      var scope = nock('https://app.getsentry.com')
+        .filteringRequestBody(/.*/, '*')
+        .post('/api/269/store/', '*')
+        .delay(10)
+        .reply(200, function (uri, body) {
+          zlib.inflate(new Buffer(body, 'base64'), function (err, dec) {
+            if (err) return done(err);
+            var msg = JSON.parse(dec.toString());
+            msg.message.should.equal('test');
+          });
+          firedCallback.should.equal(false);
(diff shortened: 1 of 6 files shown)


benvinegar:
Looks okay, but it's hard to see through the indentation.

My only ask is: a test that demonstrates that the callback behavior has changed to triggering after send (vs before send).

LewisJEllis:
@benvinegar pushed a commit for that.

test/raven.client.js
```diff
@@ -382,6 +382,39 @@ describe('raven.Client', function () {
...
+          return 'OK';
+        });
+
+      client = new raven.Client(dsn);
+
+      client.process({
+        message: 'test',
+      }, function () {
+        firedCallback = true;
+        sentResponse.should.equal(true);
```
benvinegar: How does this test guarantee that `done()` is called after the `process` callback?

LewisJEllis: Good catch, had mixed up with a change made in one of the other PRs and this wasn't checking it correctly. Fixed.

PR #219, opened 2016-10-24
Add unhandledRejection handler

LewisJEllis:


benvinegar:
I strongly feel this should be an option, and not done by default. Also because a lot of promise libraries return non-error values for "reason", which makes entries in Sentry look bad.

LewisJEllis:
Added options parameter to `#install()` so you have to do:
```
Raven.config(dsn).install({ unhandledRejection: true });
```

LewisJEllis:
Figured you might say that, but hadn't come up with a good way to deal with older nodes that don't have Promises/don't have Promises linked to `unhandledRejection`. Realized this morning that I can just `process.emit('unhandledRejection', ...)` 

Added tests.

LewisJEllis:
Yea agreed, thought right now `#install` in raven-js doesn't take a callback at all (which we do here carried over from `#patchGlobal`), and the signature/behavior of that callback are kind of wonky here, but I've not yet figured out how it should be instead.

jjinux:
Hmm, I don't think this is in the documentation.

kamilogorek:
It is https://docs.sentry.io/clients/node/usage/#promises

PR #220, opened 2016-10-24
Add shouldSendCallback, set*Callback methods

LewisJEllis:


test/raven.client.js
```diff
@@ -382,6 +382,24 @@ describe('raven.Client', function () {
...
+    it('should respect shouldSendCallback', function (done) {
+      client = new raven.Client(dsn, {
+        shouldSendCallback: function (data) {
+          return false;
+        }
+      });
+
+      // neither of these should fire, so report err to done if they do
+      client.on('logged', done);
+      client.on('error', done);
```
benvinegar: Won't this just end the test if they are invoked?

lib/client.js
```diff
@@ -309,6 +317,41 @@ extend(Raven.prototype, {
     utils.consoleAlert('setTagsContext has been deprecated and will be removed in v2.0');
     this._globalContext.tags = extend({}, this._globalContext.tags, tags);
     return this;
+  },
+
+  setCallbackHelper: function (propertyName, callback) {
+    var original = this[propertyName];
+    if (typeof callback === 'function') {
+      this[propertyName] = function (data) {
+        return callback(data, original);
```
benvinegar: Can you add a test that verifies that `original` is passed if there is a pre-existing callback?

lib/client.js
```diff
@@ -114,12 +116,18 @@ extend(Raven.prototype, {
...
+    var shouldSend = true;
+    if (!this._enabled) shouldSend = false;
+    if (this.shouldSendCallback && !this.shouldSendCallback()) shouldSend = false;
+
+    if (shouldSend) {
       this.send(kwargs, cb);
     } else {
-      setImmediate(function () {
+      // wish there was a good way to communicate to cb why we didn't send; worth considering cb api change?
+      // avoiding setImmediate here because node 0.8
```
benvinegar: Good point. Any thoughts on what this might look like?

test/raven.client.js
LewisJEllis: `done` takes an err argument so if anything is passed to it the test fails; lets you easily say "as long as this callback isn't passed an error, we're good".

lib/client.js
LewisJEllis: Added.

test/raven.client.js
```diff
@@ -382,6 +382,98 @@ describe('raven.Client', function () {
...
+
+    it('should pass original dataCallback to newer dataCallback', function (done) {
+      var scope = nock('https://app.getsentry.com')
+        .filteringRequestBody(/.*/, '*')
+        .post('/api/269/store/', '*')
+        .reply(200, function (uri, body) {
+          zlib.inflate(new Buffer(body, 'base64'), function (err, dec) {
+            if (err) return done(err);
+            var msg = JSON.parse(dec.toString());
+            msg.extra.foo.should.equal('bar');
```
benvinegar: Will this assertion / `zlib.inflate` run before `done` is called on L473?

LewisJEllis: Caught me slippin again. I'm actually not sure, could go either way depending on implementation details. Switched to `inflateSync` to be more obviously sequenced.

benvinegar:
wfm (although tests are failing)

PR #223, opened 2016-10-28
Middleware fixes

LewisJEllis:
Adds new Raven-instance-oriented middleware-returning-methods:
```javascript
var Raven = require('raven');
Raven.config(dsn).install();
...
app.use(Raven.requestHandler());
...
app.use(Raven.errorHandler());
```

Also BC-ifies old raven.middleware.express.* interface to delegate to these after ensuring existence of client instance.

/cc @MaxBittker @benvinegar


LewisJEllis:
We're not de-duping clients in the BC patchwork, only in the new way to do things, so not really.

PR #225, opened 2016-11-01
Cleaning up wrap, adding intercept

LewisJEllis:
This is a little WIPpy, need to test more, but I realized 1) usefulness of an intercept function and 2) some simplifications to wrap.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -258,16 +258,28 @@ extend(Raven.prototype, {
     return wrapped;
   },
 
-  intercept: function (cb) {
+  intercept: function (opts, func) {
     var self = this;
     var wrapped = function () {
       var err = arguments[0];
       if (err instanceof Error) {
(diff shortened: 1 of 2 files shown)


LewisJEllis:
@benvinegar take a look?

Main consideration - is `intercept` the right name for this; was originally inspired by [domain.intercept](https://nodejs.org/api/domain.html#domain_domain_intercept_callback) but this isn't really using domains, just an errback pattern; also thinking about `interceptErr` or `wrapErr` or something.

MaxBittker:
re: naming, the special thing about this intercept function is that it works with the node err callback argument convention, vs the exception world (which doesn't work with async callbacks). That being said, would a name that references this is specifically for callback errs be more descriptive?

MaxBittker:
the code lgtm

PR #236, opened 2016-11-21
node-uuid -> uuid

scttcper:
can find more information on node-uuid https://github.com/kelektiv/node-uuid/issues/116


LewisJEllis:
Looks good, thanks for bringing this to our attention.

Unfortunately for our users, all version of node-uuid have been marked as deprecated in npm, so anyone installing any version of raven-node will see a deprecation warning, which often leads to significant confusion and concern among those not well-versed in the npm ecosystem. I'll publish 0.12.3 with this change to provide a way for raven-node users to avoid this until 1.0.0 lands.

LewisJEllis:
Published 0.12.3.

PR #239, opened 2016-11-28
Deprecate/remove captureQuery

LewisJEllis:
/cc @MaxBittker @benvinegar

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -214,20 +214,6 @@ extend(Raven.prototype, {
     return eventId;
   },
 
-  captureQuery: function captureQuery(query, engine, kwargs, cb) {
-    if (!cb && typeof kwargs === 'function') {
-      cb = kwargs;
-      kwargs = {};
-    } else {
-      kwargs = kwargs || {};
-    }
-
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -409,11 +395,15 @@ extend(Raven.prototype, {
     return result;
   },
   captureError: function captureError() {
-    utils.consoleAlert('captureError has been deprecated and will be removed in v2.0');
+    utils.consoleAlert('captureError has been deprecated and will be removed in v2.0; use captureException instead');
     return this.captureException.apply(this, arguments);
   },
+  captureQuery: function captureQuery() {
+    utils.consoleAlert('captureQuery has been deprecated and will be removed in v2.0');
```
benvinegar: > will be removed

Doesn't this straight-up remove it? Why not keep the original implementation w/ this warning and remove in 2.0?

LewisJEllis:
My understanding from a discussion with Cramer was that the
server/docs don't even know what query captures are anymore, so it's safe
to remove as long as it doesn't *break* existing code. If you think we
should keep the existing functionality for now, I'm fine with that.

PR #240, opened 2016-11-28
Automatic breadcrumbs

LewisJEllis:
Usage:
```javascript
Raven.config(sentryDsn, {
  autoBreadcrumbs: {
    console: true,
    http: true,
    postgres: true
  }
}).install();
```

Todo:
- [x] Fix up configuration scheme
- [x] Become more sure that bad http requests won't explode/loop
- [x] Ensure that https requests are captured via http instrumentation
- [x] Don't capture breadcrumbs for our own requests
- [x] Don't capture breadcrumbs before instrumentation is finished
- [x] Tests (see #244)

/cc @MaxBittker @benvinegar @mitsuhiko


lib/breadcrumbs.js
```diff
@@ -0,0 +1,121 @@
...
+  http.get = function (options, cb) {
+    var req = http.request(options, cb);
+    req.end();
+    return req;
+  };
+}
+
+function wrapPostgres(Raven) {
+  // Using fill helper here is hard because of `this` binding
+  var pg = require('pg');
```
benvinegar: What if `pg` isn't available?

lib/breadcrumbs.js
```diff
@@ -0,0 +1,121 @@
...
+    });
+  };
+
+  ['debug', 'info', 'warn', 'error', 'log'].forEach(wrapConsoleMethod);
+}
+
+function wrapHttp(Raven) {
+  var http = require('http');
+  var OrigClientRequest = http.ClientRequest;
+  var ClientRequest = function (options, cb) {
```
benvinegar: Definitely need good test coverage on this.

lib/client.js
```diff
@@ -75,9 +81,19 @@ extend(Raven.prototype, {
...
 
     registerExceptionHandler(this, cb);
-    if (opts && opts.unhandledRejection) {
-      registerRejectionHandler(this, cb);
+    if (opts) {
+      if (opts.unhandledRejection) {
+        registerRejectionHandler(this, cb);
+      }
+
+      if (opts.autoBreadcrumbs) {
```
benvinegar: This should be able to work like this:

```javascript
Raven.config('your-dsn', {
  autoBreadcrumbs: true // everything
});
```

Where in a future version, `autoBreadcrumbs: true` becomes the default.

lib/breadcrumbs.js
LewisJEllis: My thinking here was "if you aren't using pg, don't do `pg: true` in the config", but I guess for the `autoBreadcrumbs: true` turn-it-all-on option we should just do a try/catch for all of the might-not-exist cases.

LewisJEllis:
Okay, makes sense - I hadn't really thought forward to when "autoBreadcrumbs: true" will be default, but yea, might as well infer/be defensive instead of relying on user explicitness.

LewisJEllis:
@benvinegar I think this should be all ready to review; did some basic tests in a PR on top of this one #244.

lib/breadcrumbs.js
```diff
@@ -0,0 +1,138 @@
...
+      Raven.captureBreadcrumb({
+        category: 'postgres',
+        message: text
+      });
+      origQuery.call(this, text);
+    };
+  }
+};
+
+function instrument(key, Raven) {
```
benvinegar: We should probably have a test for this too.

lib/client.js
```diff
@@ -357,22 +391,42 @@ extend(Raven.prototype, {
       res.sentry = eventId;
       return next(err);
     };
+  },
+
+  captureBreadcrumb: function (breadcrumb) {
+    if (!this.installed) return;
```
benvinegar: Does `installed` work like it does in raven-js? (Where you can choose not to "install" raven-js but that really just affects `window.onerror` and auto instrumentation.) 

In other words – do we need this check?

lib/client.js
```diff
@@ -357,22 +391,42 @@ extend(Raven.prototype, {
...
+
+    breadcrumb = extend({
+      timestamp: +new Date / 1000
+    }, breadcrumb);
+
+    var currCtx = this.getContext();
+    if (!currCtx.breadcrumbs) currCtx.breadcrumbs = [];
+    currCtx.breadcrumbs.push(breadcrumb);
+    if (currCtx.breadcrumbs.length > this.maxBreadcrumbs) {
```
benvinegar: So `maxBreadcrumbs` is effectively per context? Is there a concern that someone could have a *lot* of contexts, and as a result breadcrumbs end up consuming a *lot* of memory?

test/instrumentation/run.sh
```diff
@@ -0,0 +1,11 @@
+# /bin/sh
+version=`node -v`
+versionMinusV=`echo $version | cut -c 2-`
+nodeRoot="node-$versionMinusV"
```
benvinegar: Do we have a Shell scripting style guide @mattrobenolt @JTCunning?

e.g. something like https://google.github.io/styleguide/shell.xml#Naming_Conventions

test/instrumentation/run.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ ! -d $nodeRoot ]; then
+  url="https://codeload.github.com/nodejs/node/tar.gz/$version"
```
benvinegar: Doesn't really matter, but I don't think the double quotes are necessary here (or below).

lib/client.js
```diff
@@ -38,6 +43,26 @@ extend(Raven.prototype, {
...
     this.environment = options.environment || process.env.SENTRY_ENVIRONMENT || '';
 
+    var autoBreadcrumbDefaults = {
+      console: false,
+      http: false,
+      pg: false
+    };
+    this.maxBreadcrumbs = Math.max(0, Math.min(options.maxBreadcrumbs || 100, 100));
+    this.autoBreadcrumbs = extend({}, autoBreadcrumbDefaults);
+    if (typeof options.autoBreadcrumbs !== 'undefined') {
```
benvinegar: This is a _little_ confusing to grep at first glance. Maybe a quick comment:

```javascript
// autoBreadcrumbs: true // enable all the breadcrumbs
// autoBreadcrumbs: {
//   http: true // enable a single breadcrumb
// }

(Looks good though.)

File: lib/client.js (hunk 5 above)
Comment (LewisJEllis) [inline lib/client.js]: This was a spot where I thought about leaving a comment but didn't; I'll add one. The purpose here is to avoid capturing breadcrumbs halfway through instrumentation when we're not in a context yet. We do a consoleAlert for each instrumentation, so consider if console instrumentation happens first and then http instrumentation. Without this early exit condition we would try to capture a breadcrumb for the http instrumentation's consoleAlert.

File: lib/client.js (hunk 9 above)
Comment (LewisJEllis) [inline lib/client.js]: Added a comment.

File: lib/client.js (hunk 6 above)
Comment (benvinegar) [inline lib/client.js]: Per our conversation, I'd probably err on recording, say, 30 breadcrumbs per context and allow users to increase it to the max if they really want to. Better safe than sorry.

Comment (benvinegar) [inline lib/client.js]: 👍

File: test/instrumentation/http.test.js [hunk 10]
```diff
@@ -0,0 +1,54 @@
...
+
+console.log('Finished, failures: ' + failedTests.length + ', successes: ' + numSuccesses);
+
+// pipeline-flood is a harness issue with how it expects argvs
+// header response splitting is something about becoming more strict on allowed chars in headers
+// not worried about either one; they also fail without our instrumentation
+var knownFailures = [
+  'test-http-pipeline-flood.js',
+  'test-http-header-response-splitting.js'
+];
```
benvinegar: Did they fail in all versions? Or just 0.12 and earlier?

LewisJEllis: Pipeline-flood fails on all versions; it relies on argvs in a sort of one-off way to do a parent/child thing for server/client processes, so we're just not invoking it correctly. We could one-off special case it, but I'm not too worried. Response splitting just fails on 4.x. Various other tests fail on 0.10 and 0.12; I didn't try to list them all. The test setup on those older versions is a bit different.

test/instrumentation/run.sh
mattrobenolt: No, but this code looks ok to me.

If anything, I'd prefer the last command to be an `exec`, so:

```bash
exec node http.test.js $PWD/$nodeRoot
```

test/instrumentation/run.sh
```diff
@@ -0,0 +1,11 @@
+# /bin/sh
```
mattrobenolt: fwiw `bash` is generally more widely used. Not a big deal, but you'll likely come across more docs pointing to `bash` code that may not work with `sh`.

test/instrumentation/run.sh
mattrobenolt: This is true, they're not necessary, but I personally prefer them as well for readability when working with variables.

PR #243, opened 2016-12-08
Add instrumentation tests for node-http against its own tests

LewisJEllis:
/cc @benvinegar @maxbittker

This is a sort of one-off harness to run node's core http module's tests against our instrumented version of it. This gives me a good amount of confidence that our instrumentation doesn't blow anything up, but I'm still working on some raven-level "did we catch the breadcrumb off the instrumentation" type tests in addition to these.

Can you go through this readme to make sure you can run them on your system? It'll take a few minutes to clone node, but running the tests only takes ~40 seconds.

### DIFF
--- a/test/instrumentation/README.md
+++ b/test/instrumentation/README.md
@@ -1,0 +1,25 @@
+# Instrumentation tests
+
+These are tests where we run our breadcrumb instrumentation on a module, then run the instrumented module's original test suite to make sure we didn't break anything.
+
+We don't run these tests in CI; they have one-off harnesses and various requirements to run which may be tricky to provide in CI, and the results of these tests are unlikely to change without significant changes to instrumentation logic which would warrant running them manually during development.
+
+We have the following tests:
+- node core http (`node-http.test.js`)
+
+We may add in the future:
+- postgres, mysql, etc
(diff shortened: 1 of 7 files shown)


test/instrumentation/README.md
```diff
@@ -0,0 +1,25 @@
...
+- postgres, mysql, etc
+- other database drivers
+- any other node core modules we choose to instrument
+
+### Usage:
+In short, you need to have node checked out to the proper version, then run `node node-http.test.js "/Path/to/node"`.
+
+Exact steps to run, starting from the directory above `raven-node`:
+```bash
+git clone git@github.com:nodejs/node.git
```
benvinegar: We should suggest:

```bash
git clone --depth=1 git@github.com:nodejs/node.git
```

benvinegar: ^Ignore, totally didn't consider the need to check out suites from specific versions.

LewisJEllis:
@benvinegar take another look? Had to drop 0.10 and 0.12 from the instrumentation tests since the test suites changed a good bit in between v0.12 and v4, but v4+ are running and passing on Travis. I also updated the readme a bit.

PR #245, opened 2016-12-13
Update connect middleware for 1.0

dhritzkiv:
Several changes:

- Reuse of an existing client

```js
client instanceof Raven.Client
```

returned `false`, preventing an existing client from being used.

- Create raven instances with `Raven.config` instead of `new Raven.Client`

- DRYer code. Made the existing client check into a reusable function.

### DIFF
--- a/lib/middleware/connect.js
+++ b/lib/middleware/connect.js
@@ -7,7 +7,7 @@ var connectMiddleware = function (client) {
   return connectMiddleware.errorHandler(client);
 };
 
-var returnOrCreateClient = function(client) {
+var returnOrCreateClient = function (client) {
   // Use an existing client if an instance of a client, or configure a new one.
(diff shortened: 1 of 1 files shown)


LewisJEllis:
@dhritzkiv thanks for the report/fix. I see the issue here - the new recommended usage is fine, but the backwards compat interface on top of it is slightly broken.

I generally agree with most of your changes except the switch to using `Raven.config` instead of `new Raven.Client`. `Raven.config` is not an instance factory - it's a config method on an existing instance.Changing `new Raven.Client` to `Raven.config` here will just update the config on the singleton instance and return that, rather than creating a new instance with the given config.

I'm going to take your PR and make this change + a couple small tweaks on top and publish it to get the patch out quickly. This will go out as v1.0.1 and v1.1.1.

LewisJEllis:
Resolved in #246.

PR #246, opened 2016-12-14
Fix middleware backwards compatibility

LewisJEllis:
- Changed to use `instanceof Raven.constructor` instead of `instanceof Raven.client`
- Added deprecation warnings; these toplevel functions are the old way to do it and people should have their own instances and use the instance methods instead, rather than having to pass in an instance or let one be created for them.

Based on #245 from @dhritzkiv. Will merge into master and also cherry pick into a `releases/1.0.x` branch, then publish 1.0.1 and 1.1.1.

/cc @benvinegar

### DIFF
--- a/lib/middleware/connect.js
+++ b/lib/middleware/connect.js
@@ -16,15 +16,18 @@ var getClient = function (clientOrDSN) {
 // Error handler. This should be the last item listed in middleware, but
 // before any other error handlers.
 connectMiddleware.errorHandler = function (clientOrDSN) {
-  utils.consoleAlert('top-level Raven.middleware.errorHandler has been deprecated and will be removed in v2.0; use Raven.errorHandler() instance method instead');
+  utils.consoleAlert('top-level Raven.middleware.*.errorHandler has been deprecated and will be removed in v2.0; use Raven.errorHandler() instance method instead');
   return getClient(clientOrDSN).errorHandler();
 };
 
 // Ensures asynchronous exceptions are routed to the errorHandler. This
 // should be the **first** item listed in middleware.
 connectMiddleware.requestHandler = function (clientOrDSN) {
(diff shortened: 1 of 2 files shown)


lib/middleware/connect.js
```diff
@@ -1,24 +1,30 @@
...
+  return clientOrDSN instanceof Raven.constructor ? clientOrDSN : new Raven.Client(clientOrDSN);
+};
+
 // Error handler. This should be the last item listed in middleware, but
 // before any other error handlers.
-connectMiddleware.errorHandler = function (client) {
-  client = client instanceof Raven.Client ? client : new Raven.Client(client);
-  return client.errorHandler();
+connectMiddleware.errorHandler = function (clientOrDSN) {
+  utils.consoleAlert('top-level Raven.middleware.* methods have been deprecated and will be removed in v2.0; use instance methods instead');
```
benvinegar: "use instance methods instead"

Are people going to know what this is? Can we link to something here, or say the actual code?

Also, what's the harm in keeping this around?

LewisJEllis: Sort of trying to keep that warning from being excessively long, but I'll be more specific about the methods I'm referring to.

Harm is that a user might just pass their DSN string in 3 places and end up with 3 instances, two of which are sort of hidden and probably entirely unnecessary, when they really just want one instance. Might result in situations where e.g. user expects to get breadcrumbs but doesn't. I want to avoid multiple instances entirely unless the user is explicitly asking for them and knows what they're doing.

benvinegar: Good explanation – I agree. My fear is that people will see that message and have no idea what you're talking about, especially since it's not yet reflected in docs.

Can we do something like:

```javascript
utils.consoleAlert('top-level Raven.middleware.* methods have been deprecated and will be removed in v2.0; use client.requestHandler()');
```

LewisJEllis:
Published 1.0.1 and 1.1.1; thanks @dhritzkiv!

PR #251, opened 2016-12-23
Pass kwargs into shouldSendCallback() as param

jeffwhelpley:
In the `shouldSendCallback` function, we need the args in order to determine whether we want to send to sentry based on the current data. This is exactly how things work for the JavaScript client version of raven.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -182,7 +182,7 @@ extend(Raven.prototype, {
 
     var shouldSend = true;
     if (!this._enabled) shouldSend = false;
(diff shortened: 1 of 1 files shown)


LewisJEllis:
Whoops, good catch, thanks!

jeffwhelpley:
Awesome. Can you publish a new version of the library to npm?

LewisJEllis:
@jeffwhelpley published in 1.1.2, sorry for the delay!

PR #268, opened 2017-02-07
Pr/118 parseStack respects windows paths 

MaxBittker:
I tested this on windows and osx and everything seems to be working correctly! thank you @stambata and @seanbriceland  original PR: https://github.com/getsentry/raven-node/pull/118

### DIFF
--- a/test/raven.utils.js
+++ b/test/raven.utils.js
@@ -102,10 +102,15 @@ describe('raven.utils', function () {
       var parseStack = raven.utils.parseStack;
       var callback = function (frames) {
         var frame = frames.pop();
+
+        /* verify that the frame has the properties: pre_context, context_line
+        and post_context, which are valid for in-app stack lines only.*/
         frame.pre_context.should.be.an.instanceOf(Array);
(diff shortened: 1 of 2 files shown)


test/raven.utils.js
```diff
@@ -97,6 +97,45 @@ describe('raven.utils', function () {
...
+
+    it('should extract context from last stack line', function (done) {
+      var parseStack = raven.utils.parseStack;
+      var callback = function (frames) {
+        var frame = frames.pop();
+        frame.pre_context.should.be.an.instanceOf(Array);
+        frame.context_line.should.be.type('string');
+        frame.context_line.trim().should.endWith('undeclared_function();');
+        frame.post_context.should.be.an.instanceOf(Array);
+        done();
```
LewisJEllis: Can we add an assertion here on what `frame.in_app` should be, and add a comment to the effect of part 4 of [this comment](https://github.com/getsentry/raven-node/pull/118#r31550565) explaining what's going on in this test? It's not initially obvious that this tests anything not existing before this PR.

Issue #266, opened 2017-02-06
Context lost and request hangs when uncaughtException happens


rwky:
I noticed when using express some errors were missing context (i.e. the url and IP).

After some investigation I've found it's due to uncaughtExceptions, this happens on any POST request that uses the body-parser middleware.

I've narrowed it down to a small test case 

```js
var app = require('express')()
var Raven = require('raven')

Raven.config('config_url').install()
app.use(Raven.requestHandler())

app.use(function(req, res, next) { 
    req.on('data', function() {})
    req.on('end', function(){ next() })
})

app.post('/broken', function (req, res) {
    process.nextTick(function(){
        throw new Error('Broken!')
    })
})

app.post('/works', function(req, res) {
    throw new Error('Works!')
});

app.use(Raven.errorHandler())

app.use(function (err, req, res, next) {
    res.statusCode = 500
    console.log('Sentry ' + res.sentry)
    res.end('Sentry ' + res.sentry + '\n')
})

app.listen(4000)

```
If you run `curl http://localhost:4000/works -d 'test'` it'll behave as expected, if you run `curl http://localhost:4000/broken -d 'test'` the reported error will miss the context **and** the request won't complete since express error handler is never called.

Due to the combination of the `req.on('end', ...` listener and the `process.nextTick` `/broken` generates and `uncaughtException`

Ideally I'd want sentry to log the exception with context, call the error handler so the request ends and then allow me to cleanup (using the callback to install()) which could then terminate the process gracefully.

Does this sound possible?

LewisJEllis:
> Due to the combination of the req.on('end', ... listener and the process.nextTick /broken generates and uncaughtException

Aha! What's happening here is kinda crafty, but you've given nearly the perfect repro case to point me at the issue. Thanks for the great report @rwky.

This is all just an initial look and I haven't verified the repro/potential fix yet, but your repro and description of observed behavior are very consistent are very consistent with what I think is going on. This indeed will only happen when `next` (or some other continuation) is invoked by an event listener, and further only when that event listener is on an event emitter that existed *before* the request-scoped domain (which our express middleware uses to associate context) is created. In this case, we have the continuation `next` invoked by a listener attached to `req` which existed before our `requestHandler` middleware creates the context domain.

[...]

PR #269, opened 2017-02-08
Explicitly bind req/res to context domain in express requestHandler

LewisJEllis:
Managed to repro/test that this fixes issue in #266. The test case is a little tricky but hopefully makes sense.

/cc @benvinegar @MaxBittker @rwky

### DIFF
--- a/test/raven.client.js
+++ b/test/raven.client.js
@@ -1018,4 +1018,29 @@ describe('raven.middleware', function () {
     client2.should.not.equal(client1);
     client2.should.be.an.instanceof(raven.constructor);
   });
+
+  it('should explicitly add req and res to the domain', function (done) {
+    var client = new raven.Client(dsn).install();
+
+    var EventEmitter = require('events');
+    var e = new EventEmitter();
+    e.on('done', function () {
+      // Context won't propagate here without the explicit binding of req/res done in the middleware
+      setTimeout(function () {
+        client.getContext().breadcrumbs.length.should.equal(1);
+        done();
+      }, 0);
+    });
+
+    // Pass e as the req/res, so e will be added to the domain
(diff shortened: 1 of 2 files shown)


rwky:
I noticed you put a fix in for node 0.10, this and 0.12 aren't supported by the node core team https://github.com/nodejs/LTS no harm in adding in this case since it's just a test case but in general might not be worth the effort of supporting node < v4

LewisJEllis:
Yea, we're certainly not holding back any new development/capabilities for the sake of node < v4, but we know we have a number of users still on those older versions (despite node's guidelines/pleading), so we support them as long as long as it's not a blocker/overly obnoxious to do so (which thus far it hasn't been).

The plan is roughly that when async_hooks shows up in node core, we'll do a new major version that supports only versions of node having async_hooks, and keep the older major version maintenance-only for everything pre-async_hooks.

test/raven.client.js
```diff
@@ -1018,4 +1018,31 @@ describe('raven.middleware', function () {
...
+    var client = new raven.Client(dsn).install();
+
+    var EventEmitter = require('events');
+    // node 0.10 compat
+    if (process.version <= 'v0.11') EventEmitter = EventEmitter.EventEmitter;
+    var e = new EventEmitter();
+    e.on('done', function () {
+      // Context won't propagate here without the explicit binding of req/res done in the middleware
+      setTimeout(function () {
+        client.getContext().breadcrumbs.length.should.equal(1);
```
MaxBittker: I assume that breadcrumbs[0].mesage==='test breadcrumb' here? is there a chance that some other one got added and this doesn't work?

LewisJEllis: Yea, good call, I added an explicit check for that.

Issue #231, opened 2016-11-11
req.user from Sequelize causes server to hang


idris:
When an error is picked up by the connect errorHandler middleware, if that request contains a `user` object from Sequelize, the server hangs for a long time.

I assume this happens because Raven is trying to serialize the User object, and that gets into all kinds of weird things, including the Postgres driver itself.

Side note: This also causes #210 and #212, because the Postgres driver has a bug where if you enumerate all properties, it will try to require `pg-native`, even if it's not in your dependencies.

### Proposed Solution
There are a few ways to resolve this.  I prefer the first.
#### Request sanitizing function
Allow a function to be passed into the middleware that accepts a request and returns a sanitized request.  My implementation would look like this:
```
(req) => {
  const { user, ...sanitizedRequest } = req; // Remove user from req
  sanitizedRequest.user = { id: user.id }; // Add userId for context
  return sanitizedRequest;
}
```

#### Recognize Sequelize models in req.user
Since req.user is popularly used with Sequelize, in `parseRequest` we could recognize that it's a sequelize model and call `req.user.get({ plain: true })` to make it a plain JS object.

LewisJEllis:
Awesome report/breakdown @idris - thanks! This brings a lot of clarity around that `pg-native` issue. I'm definitely on board with fixing this in one way or another; Sequelize is way too popular for us to not work with. For now, the closest thing to a workaround is rolling your own slightly-modified-but-mostly-just-duplicate of the error handler middleware to call `parsers.parseRequest(req, { user: req.user.get({ plain: true }));` instead of `parsers.parseRequest(req);`, and that's pretty rough.

In general I think it's better to avoid ecosystem-member-interop-specific code paths like what you describe in option 2, so I would be inclined to agree with you on the request sanitizing function as the flexible solution (even though option 2 does make a simpler UX in this case). I think that function would fit better as a configurable option on the Raven instance rather than an option to the middleware, though, and I still need to figure out how user context data should fit in vs user data off the req obj, which one should take precedence, should they merge or just either-or, etc, but that isn't really the core issue and I think those details will fall in to place.

The other point of consideration is that with upcoming breadcrumb support, we're going to have to have a fair amount of ecosystem-player-specific code for automatically capturing queries, logs, etc from various popular libraries. I'm still getting my thoughts in line on that, but it makes me wonder if maybe library-interop-specific code isn't such a bad thing.

[...]

Issue #82, opened 2014-07-23
Additional APIs / improvements to parsing


rcoup:
Currently it's actually pretty messy to add a contextual description to an error, or add parsed request info to messages or errors outside of the middleware. 

``` javascript
function someView(req, res) {
  doStuff(function(err, data) {
    if (err) {
      parsers.parseError(err, parsers.parseRequest(req), function(kwargs) {
        var result = client.captureMessage("error/context description", kwargs);
        res.send(400, "Badness: " + client.getIdent(result));
      });
    } else {
      res.send(200, data);
    }
  }
}
```

Seems we could make it a lot simpler pretty easily:
1. add an optional description to captureError():

```
client.captureError(Error error[, String description][, Object options][, Function callback])
```
1. automatically look for a `request` (or `req`) in any `options` passed to captureMessage() and captureError(). If found, run parseRequest() on it.

The above then becomes

``` javascript
function someView(req, res) {
  doStuff(function(err, data) {
    if (err) {
      client.captureError(err, "context description", {request: req}, function(result) {
        res.send(400, "Badness: " + client.getIdent(result));
      });
    } else {
      res.send(200, data);
    }
  }
}
```

and passing request information along with a captureMessage() call becomes:

``` javascript
client.captureMessage("some message", {request:req});
```

Thoughts on the concept?

benvinegar:
@MaxBittker – this is an old issue, but I've seen a bunch of related issues more recently. I'm in favor of us automatically attempting parsing a `request` property if we find it as an option to `captureException`/`captureMessage`.

LewisJEllis:
Couple more related areas we can grab the request more automatically: in `process` if there's a `context.request` value and no `kwargs.request`, we might as well read the request from the context. Then in the `requestHandler` middleware we can add the request to the context; this way errors captured after the `requestHandler` middleware but *not* by the `errorHandler` middleware (i.e. if someone does a manual `captureException`) will still have the request associated, when they currently do not.

I think there are a couple more similar "could be slightly more automatic with context association for semi-manual capture cases" spots but I don't remember them exactly right now - will make sure to shore up any gaps.

LewisJEllis:
Resolved in #310

PR #274, opened 2017-02-22
Adds parseUser option to control user parsing behavior

LewisJEllis:
Should address #231. Will also add some flexibility for both #82 and another slight improvement that we can make in the requestHandler middleware for automatically-grabbing-request.

/cc @benvinegar @MaxBittker and somewhat refactors request parsing so /cc @mattrobenolt.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -73,6 +73,29 @@ Those configuration options are documented below:
             extra: {planet: {name: 'Earth'}}
         }
 
+.. describe:: parseUser
+
+    Controls how Raven tries to parse user context when parsing a request object.
+
+    An array of strings will serve as a whitelist for fields to grab from ``req.user``.
+    ``true`` will collect all keys from ``req.user``. ``false`` will collect nothing.
+
+    Defaults to ``['id', 'username', 'email']``.
+
+    Alternatively, a function can be provided for fully custom parsing:
+
+    .. code-block:: javascript
+
+        {
+            parseUser: function (req) {
+                // custom user parsing logic
+                return {
+                    username: req.specialUserField.username,
+                    id: req.specialUserField.getId()
+                };
+            }
+        }
+
 .. describe:: dataCallback
 
     A function that allows mutation of the data payload right before being
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -53,74 +53,48 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
   });
 };
 
-module.exports.parseRequest = function parseRequest(req, kwargs) {
-  kwargs = kwargs || {};
+module.exports.parseRequest = function parseRequest(req, parseUser) {
+  var kwargs = {};
 
   // headers:
-  //
-  //   node: req.headers
-  //   express: req.headers
+  //   node, express: req.headers
   //   koa: req.header
-  //
   var headers = req.headers || req.header || {};
 
   // method:
-  //
-  //   node: req.method
-  //   express: req.method
-  //   koa: req.method
-  //
+  //   node, express, koa: req.method
   var method = req.method;
 
   // host:
-  //
-  //   node: req.headers.host
   //   express: req.hostname in > 4 and req.host in < 4
   //   koa: req.host
-  //
+  //   node: req.headers.host
   var host = req.hostname || req.host || headers.host || '<no host>';
 
   // protocol:
-  //
   //   node: <n/a>
-  //   express: req.protocol
-  //   koa: req.protocol
-  //
+  //   express, koa: req.protocol
   var protocol = req.protocol === 'https' || req.secure || (req.socket || {}).encrypted ? 'https' : 'http';
 
   // url (including path and query string):
-  //
-  //   node: req.originalUrl
-  //   express: req.originalUrl
+  //   node, express: req.originalUrl
   //   koa: req.url
-  //
   var originalUrl = req.originalUrl || req.url;
 
   // absolute url
-  var url = protocol + '://' + host + originalUrl;
+  var absoluteUrl = protocol + '://' + host + originalUrl;
 
-  // query string
-  //
+  // query string:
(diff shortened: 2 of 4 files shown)


test/raven.parsers.js
```diff
@@ -416,38 +416,80 @@ describe('raven.parsers', function () {
...
-          username: 'janedoe',
-          email: 'hello@janedoe.com',
-          ip_address: '127.0.0.1'
+        it('should limit to default whitelisted fields', function () {
+          var parsed = raven.parsers.parseRequest(mockReq);
+          parsed.should.have.property('user', {
+            username: 'janedoe',
+            email: 'hello@janedoe.com',
+            ip_address: '127.0.0.1'
+          });
```
MaxBittker: does this check that random is -not- set?

docs/config.rst
```diff
@@ -73,6 +73,29 @@ Those configuration options are documented below:
...
+.. describe:: parseUser
+
+    Controls how Raven tries to parse user context when parsing a request object.
+
+    An array of strings will serve as a whitelist for fields to grab from ``req.user``.
+    ``true`` will collect all keys from ``req.user``. ``false`` will collect nothing.
+
+    Defaults to ``['id', 'username', 'email']``.
+
+    A function can be provided for custom parsing:
```
MaxBittker: maybe, "alternatively, a function ..."

test/raven.parsers.js
LewisJEllis: yes, it checks that the user property has that exact value

docs/config.rst
LewisJEllis: 👍

lib/parsers.js
```diff
@@ -53,123 +53,105 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
...
+          if ({}.hasOwnProperty.call(req.user, key)) {
+            user[key] = req.user[key];
+          }
         }
+      } else {
+        parseUser.forEach(function (fieldName) {
+          if ({}.hasOwnProperty.call(req.user, fieldName)) {
+            user[fieldName] = req.user[fieldName];
+          }
+        });
```
MaxBittker: this looks good to me but this is a big gnarly block of logic. consider labeling some of the control flow paths?

test/raven.parsers.js
MaxBittker: thanks

lib/parsers.js
```diff
@@ -53,123 +53,105 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
...
+  //   node, express, koa: req.body
   var data = req.body;
   if (['GET', 'HEAD'].indexOf(method) === -1) {
     if (typeof data === 'undefined') {
       data = '<unavailable>';
     }
   }
 
-  if (data && {}.toString.call(data) !== '[object String]') {
+  if (data && typeof data !== 'string' && {}.toString.call(data) !== '[object String]') {
```
benvinegar: Why add this?

lib/parsers.js
```diff
@@ -53,123 +53,105 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
   });
 };
 
-module.exports.parseRequest = function parseRequest(req, kwargs) {
-  kwargs = kwargs || {};
+module.exports.parseRequest = function parseRequest(req, parseUser) {
```
benvinegar: Is `parseRequest` a public/documented API method? If so, might want to consider how to make this signature more extendable going forward.

e.g. I'd almost prefer if we kept `kwargs` and you did:

```javascript
parseRequest(req, {
  parseUser: ['username']
});
```

lib/parsers.js
LewisJEllis: ehh...normally id agree but i did it that way at first and it felt like borderline overcommenting since each branch has a pretty straightforward condition check, so i made sure to describe the 5 cases together above instead

lib/parsers.js
```diff
@@ -53,123 +53,105 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
...
+  //   absent: grab only id, username, email from req.user
+  //   false: capture nothing
+  //   true: capture all keys from req.user
+  //   array: provided whitelisted keys to grab from req.user
+  //   function :: req -> user: custom parsing function
+  if (parseUser == null) parseUser = ['id', 'username', 'email'];
+  if (parseUser) {
+    var user = {};
+    if (typeof parseUser === 'function') {
+      user = parseUser(req);
```
benvinegar: Is there a reason we pass `req` to `parseUser` and not just strictly the `user`?

(I mean, this is probably more flexible, but want to get your take.)

lib/parsers.js
LewisJEllis: Not public; only callsite is currently in `errorHandler` middleware, soon we'll add another in `requestHandler` and in `process` but all internal. The `kwargs` param was unused so I got rid of it.

lib/parsers.js
LewisJEllis: In my mind the function option exists primarily as a catch-all for "well what if it's not at `req.user`..."; other main case would be if we want to pull in other stuff off the req object (like maybe we want stuff from `req.session`), so either way I pass the entire `req` object for max flexibility.

lib/parsers.js
LewisJEllis: The `{}.toString.call` check is really checking two cases without being obvious about it, and I sort of prefer using `typeof` as much as possible and avoiding checking object toString types when we don't need to be. This better delineates/escalates through the cases we care about IMO (and `typeof` is a cheaper check in the common case, not that we care about a trivial perf diff in a once-per-error check).

lib/parsers.js
benvinegar: k

lib/parsers.js
benvinegar: 👍

Issue #273, opened 2017-02-18
Auto-breadcrumbs break some aws-sdk API calls


ide:
When Raven's autoBreadcrumbs option overrides Node's `http` module, it adds behavior that adds a `data` listener to HTTP response streams. This can cause issues because Node's readable streams start out in "paused" mode by default and adding a `data` listener (as well as other conditions described here: https://nodejs.org/api/stream.html#stream_two_modes) put it in "flowing" mode.

The aws-sdk package expects HTTP response streams to be in paused mode. If the stream gets put into flowing mode, aws-sdk doesn't have the right event listener and ignores data from the stream, causing it to get a null response and return an empty object to the user. See https://github.com/aws/aws-sdk-js/issues/1365 for more info.

LewisJEllis:
Hey @ide - thanks for the report. This also came up in #248; I suspected it could be poor interaction with our `data` listener [here](https://github.com/getsentry/raven-node/blob/master/lib/breadcrumbs.js#L103), but hadn't dug into it yet. 

The spot in the AWS SDK that you linked makes it seem pretty clear that that's what's going on. I should be able to change the http instrumentation to grab response details via overriding `ClientRequest.emit` instead of actually attaching a `response` event listener. The change is pretty straightforward, I just need to figure out a good way to test it.

LewisJEllis:
This was closed back in #276 and has been publishes to npm.

PR #276, opened 2017-02-23
Make http instrumentation use req.emit instead of response event handler

### LINKED ISSUE #248: Auto breadcrumbs HTTP causes DynamoDB errors in AWS Lambda environments
I tried using `autoBreadcrumbs: true` in an AWS Lambda function, but when deployed to AWS all my DynamoDB operations started failing with `Error: CRC32CheckFailed: CRC32 integrity check failed` errors.

The errors went away after changing my Raven config to `autoBreadcrumbs: { console: true }`.

LewisJEllis:
Should fix #248 and #273. Now our instrumentation for http breadcrumbs will pick up response details via monkeypatching `ClientRequest.emit` and watching the `response` event, rather than just adding a listener for the `response` event (which then forces us to sort-of-consume the response body to avoid eating memory, which then interacts poorly with at least aws-sdk). Went down a few of rabbit holes on this but it turned out to be pretty straightforward in the end.

I upgraded our should.js devdep while trying to make a test fail more nicely, but it didn't help and I decided I don't care about tests failing nicely since they're meant to pass; left it since it also revealed a small inconsistency in our tests.

/cc @MaxBittker @benvinegar @ide


lib/breadcrumbs.js
```diff
@@ -87,21 +87,22 @@ var wrappers = {
...
+                  url: url,
+                  status_code: maybeResp.statusCode
+                }
+              });
             }
-          });
-
-          // No-op to consume response data since we added a response handler
-          // see https://nodejs.org/api/http.html#http_class_http_clientrequest
-          response.on('data', function () {});
```
LewisJEllis: This `data` listener was causing `_readableState.flowing` to be true; see https://nodejs.org/api/stream.html#stream_three_states

LewisJEllis:
Whoops, meant to reference #273.

Issue #254, opened 2017-01-01
Cannot resolve module 'pg'


udisun:
as of version 1.1.0
you have added an inline `require('pg')` on line 132 of `raven/lib/breadcrumbs.js`
this should be a dependency in package.json.
the error:
```
ERROR in ./~/raven/lib/breadcrumbs.js
Module not found: Error: Cannot resolve module 'pg' in project_dir/node_modules/raven/lib
 @ ./~/raven/lib/breadcrumbs.js 132:13-26
```

LewisJEllis:
Where do you run into this error? Do you have a snippet I can run to reproduce it? We don't depend on/require `pg`, we just look to see if the project using raven-node is using `pg` themselves and, if so, instrument the `pg` module to capture breadcrumbs. The line you point to is only run inside a try/catch, so I wouldn't expect the error to escape.

holmari:
This happens when you build with webpack. Something like this helps, in webpack.config.js:

```
module.exports = {
  ...
  externals: [
    {pg: true}
  ]
  node: {
    fs: 'empty'
  }
  ...
};
```

LewisJEllis:
Ah, okay, that makes sense. I didn't really think about webpack in the Node case, but there are some legitimate uses, so it would probably be better here to check for the presence of the module using `require.resolve` rather than just attempting to load it. I'm guessing that'll make webpack quit looking for it when it's not there, but I'll have to verify.

shubhamarora:
@LewisJEllis I am facing the same problem and there is no such `pg` param in my webpack config. Do we have any solution for this?

benvinegar:
@shubhamarora – are you building your js files somehow? like, with Webpack + Babel? and are you intending to use raven-node or raven-js (browser JavaScript)?

shubhamarora:
@benvinegar Yes, I am building/bundling using Weback + Babel and I am using raven-node.

CBlaize:
Same issue for me as well.  I am bundling raven-node with Webpack for use in the browser.

mattrobenolt:
> I am bundling raven-node with Webpack for use in the browser.

You shouldn't be using raven-node for that. raven-node will not run in a browser and you must use raven-js.

LewisJEllis:
[...]

PR #277, opened 2017-02-23
Add alert about node vs raven-js when it seems like a browser env

LewisJEllis:
As per comment [here](https://github.com/getsentry/raven-node/issues/254#issuecomment-281555668). I do this check in `config` so that a user can do `Raven.disableConsoleAlerts()` before `Raven.config` to avoid seeing this alert.

/cc @benvinegar - does this seem like a reasonable message and condition for "seems like a browser?"

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -23,6 +23,11 @@ nodeUtil.inherits(Raven, events.EventEmitter);
 
 extend(Raven.prototype, {
   config: function config(dsn, options) {
+    // We get lots of users using raven-node when they want raven-js, hence this warning if it seems like a browser
+    if (typeof window !== 'undefined' && typeof document !== 'undefined' && typeof navigator !== 'undefined') {
+      utils.consoleAlert('This looks like a browser environment; are you sure you don\'t want Raven.js for browser JavaScript? https://sentry.io/for/javascript');
+    }
+
     if (arguments.length === 0) {
       // no arguments, use default from environment
       dsn = process.env.SENTRY_DSN;


benvinegar:
Seems reasonable. Is there a way to disable the message? Will people want to, or care?

I think it could be shortened:

> This looks like a browser environment. Are you sure you don't want Raven.js (browser JavaScript)?

I might also prefix the output with `raven-node:` to be clear which library is outputting this.

LewisJEllis:
> Is there a way to disable the message?

Yes:

> a user can do `Raven.disableConsoleAlerts()` before `Raven.config` to avoid seeing this alert.

We already prefix it like "raven@1.1.2: <message>" - could be more clear to do raven-node but then it's not the npm package name, so I explicitly said "Node.js" in the message, but it's probably sufficiently clear anyway. I cleaned up the wording a little more in line with your suggestion.

shubhamarora:
@LewisJEllis @benvinegar As per my understanding, this error is comming while bundling. 

So let me try to explain a bit more. So when webpack comes into action. It makes a dependency tree and then it resolves and inlcude those dependencies in a bundle (which is called output in terms of webpack). This error is being thrown when webpack is bundling, so due to wrong doing, raven-node is comming onto the dependency tree and webpack tries to resolve and include it to the bundle (which should not happen as raven-node should never be shipped in a bundle to client side/browser) and it further tries to resolve the dependency (pg) of a dependency (raven-node), where it breaks.

This solution - https://github.com/getsentry/raven-node/issues/254#issuecomment-272050475 works, because this way we are telling webpack to not to lookup and resolve the dependency ('pg') (more info [here](https://webpack.github.io/docs/library-and-externals.html) but unknowingly raven-node might be getting shipped with client side bundle as far as i can think of it. IMO i dont think this fix will solve this problem. Let me know if i can help in digging this further.

Apart from this, Does typeof window check is not fair enough to check if its browser?

PR #282, opened 2017-02-28
add typings to the module

### LINKED ISSUE #189: Add TypeScript type definitions
The `raven-js` library for browsers got TypeScript definitions in https://github.com/getsentry/raven-js/pull/610, but those definitions cannot be used for the nodejs library.

The definitions should be implemented for this version as well.

netroy:
fixes #189

borrows heavily from https://github.com/getsentry/raven-node/issues/189#issuecomment-241355970

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,56 @@
+/// <reference types="node" />
+
+import { IncomingMessage, ServerResponse } from 'http'
+
+declare namespace raven {
+  export interface UserData {
+    id: string
+    handle?: string
+  }
+
+  interface CaptureCallback {
+    (result: { string: any }): void
+  }
+
+  interface DataCallback {
+    (data: { string: any }): void
+  }
+
+  interface TransportCallback {
+    (options: { string: any }): void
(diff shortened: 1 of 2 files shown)


benvinegar:
While I appreciate the effort that went into this, we're intentionally avoiding having TypeScript typings in raven-node. They're a burden to maintain, people get upset if they don't work, they _will_ be continuously out of date, and there hasn't been enough community push for them. We only support them in raven-js because Angular 2 is an officially supported platform, and it uses TypeScript by default.

We'd prefer if you consider submitting these to [DefinitelyTyped](https://github.com/definitelytyped/definitelytyped) instead.

julien-c:
Would love this as well

edvald:
This would be super useful. Pushing this to DefinitelyTyped just adds a layer of complexity and doesn't help at all with compatibility issues, imo.

netroy:
typings for raven-node landed here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/15429

PR #283, opened 2017-03-01
execute .install() cb even when dsn is not supplied

esetnik:
When installing global exception handler and disabling sending the uncaught exception handler is not called. For example:

```
// pass null to config() to disable reporting (for local development)
const ravenClient = Raven.config(null).install(function(logged, err) {
  log.error('process is terminating due to uncaught exception', err);
  process.exit(1);
});
```

This pull request causes the install callback to be populated with the values `(false, err)` when sentry reporting is disabled and a global exception is encountered.


LewisJEllis:
This has been addressed by the re-work of the `install` fatal error callback in #308.

Issue #214, opened 2016-10-16
error events unreliable properties


pekeler:
The error event does not always contain reason or statusCode as the documentation describes.

Consider this simple code to run into api rate limiting.

``` JavaScript
'use strict'
const raven = require('raven')
const ravenClient = new raven.Client(secret)

ravenClient.on('error', function(error) {
  console.error(error.reason, error.statusCode);
})

for (let count = 0; count < 100; count += 1) {
  ravenClient.captureMessage('just testing')
}
```

Sometimes, I get output as the following.

```
undefined undefined
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
undefined 429
Creation of this event was denied due to rate limiting 429
undefined 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
undefined undefined
undefined undefined
undefined undefined
undefined undefined
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
undefined undefined
undefined undefined
Creation of this event was denied due to rate limiting 429
undefined undefined
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
undefined undefined
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
undefined undefined
undefined undefined
undefined undefined
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
Creation of this event was denied due to rate limiting 429
[...]

PR #284, opened 2017-03-05
transport: enable keep-alives and limit number of sockets

mattrobenolt:


mattrobenolt:
I need to confirm yet this actually does what I think it does.

lib/transports.js
```diff
@@ -6,12 +6,16 @@ var util = require('util');
 function Transport() {}
 util.inherits(Transport, events.EventEmitter);
 
+var agentOptions = { keepalive: true, maxSockets: 100 };
```
LewisJEllis: 100 seems high to me? Also, should we provide a `maxFreeSockets` option? (https://nodejs.org/api/http.html#http_new_agent_options)

lib/transports.js
```diff
@@ -6,12 +6,16 @@ var util = require('util');
...
+
 var http = require('http');
+var httpAgent = new http.Agent(agentOptions);
 
 function HTTPTransport(options) {
   this.defaultPort = 80;
   this.transport = http;
   this.options = options || {};
+  this.agent = httpAgent;
```
LewisJEllis: We could theoretically enable users to change agent options via something like:
```javascript
    Raven.config('___DSN___', {
      transport: new raven.transports.HTTPSTransport({agentOpts: { ... }})
    });
```
but I'm not sure whether we need to or should; are there cases where users might want to do this (or we might want them to)? Guessing you have more background on our intentions behind transports being pluggable/user-configurable.

benvinegar:
*ping*

lib/transports.js
kamilogorek: Number of `maxFreeSockets` won't go over `maxSockets` number and it defaults to `256` so we should be good here.

lib/transports.js
kamilogorek: Agree on enabling user to change those settings.

kamilogorek:
Ping @mattrobenolt

lib/transports.js
kamilogorek: + adding a comment in the docs about it

mattrobenolt:
> I need to confirm yet this actually does what I think it does.

It does. This would be nice to have.

kamilogorek:
@benvinegar @mattrobenolt updated. First two commits come from https://github.com/getsentry/raven-node/pull/369 as it's related, and it should be merged first.

kamilogorek:
Ping

kamilogorek:
I just realised that this code has been partially broken due to incorrect socket name being constructed. Fixed and updated test.

This should also fix https://github.com/getsentry/raven-node/issues/214 enough for it to be closed.

PR #292, opened 2017-03-10
Add sampleRate config option

LewisJEllis:
Basically the same thing as getsentry/raven-js#885

/cc @benvinegar @dcramer

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -96,6 +96,17 @@ Those configuration options are documented below:
             }
         }
 
+.. describe:: sampleRate
+
+    A sampling rate to apply to events. A value of 0.0 will send no events,
+    and a value of 1.0 will send all events (default).
+
+    .. code-block:: javascript
+
+        {
+            sampleRate: 0.5 // send 50% of events, drop the other half
+        }
+
 .. describe:: dataCallback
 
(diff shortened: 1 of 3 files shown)


LewisJEllis:
Added tests. Feels kinda messy compared to the raven-js parallels with the nice sinon setup :/

MarkHerhold:
Sorry for the dumb question, but what is the use-case for this?

benvinegar:
@MarkHerhold – This is analogous to rate-limiting, except the rate-limiting occurs on the client, not the server. This has the upside of sending less traffic to sentry.io ($) / your own Sentry server at the expense of losing data.

Sampling is often necessary for very high traffic applications.

MarkHerhold:
Ah, so this would be for applications that are utilizing Sentry as a logging service rather than an error tracker?

benvinegar:
It's more like – if you have 1 million active users, and they all hit the same error, do you really need 1 million error reports to be sent? Probably not.

So, if you have an app with a really huge active user count, you can probably safely suppress a fraction of error reports using `sampleRate` to save bandwidth, $, etc.

MarkHerhold:
Got it, thanks!

PR #296, opened 2017-03-16
Fix autoBreadcrumb http instrumentation memory leak

### LINKED ISSUE #295: Massive memory leak when auto breadcrumbs is enabled
Unfortunately I will not be able to dig into this for you, but it should be reported so that others who can have a starting point and realize they are not alone.

I had to disable auto breadcrumbs due to the fact it was consistently consuming gigs of memory and crashing my Node process within minutes of running in production. (Roughly every twenty minutes or so it would exhaust available memory and restart the process.)

Disabling auto breadcrumbs fixed the issue.

This is on Node v7.4.0 and raven-node 1.1.4. Happens on macOS as well as whatever Docker node:7.4 is running.

Best of luck fixing this...it's a nasty one.

LewisJEllis:
PR #276/version 1.1.3 introduced a memory leak in our http instrumentation as reported in #295. The core issue was that we were monkeypatching `http.ClientRequest.prototype.emit` during every `ClientRequest` instantiation, instead of just during initial instrumentation, which meant that for each outbound http request made, we kept a bunch of stuff around in a closure.

Users on raven-node 1.1.3 or 1.1.4 with long-running processes that made outbound http requests found their memory usage growing steadily over time.

This PR refactors the `ClientRequest` instrumentation to only monkeypatch `http.ClientRequest.prototype.emit` once during initial Raven setup when we also monkeypatch the constructor. The monkeypatched `emit` method previously found the request url by closure reference, but now the constructor attaches a `__ravenBreadcrumbUrl` property to the request object and `emit` grabs that instead.

I confirmed the issue and fix using [memwatch-next](https://www.npmjs.com/package/memwatch-next) (plus express, some outbound requesting, and some apachebenching) and have ideas for how to use that to automate some "did we intro a memory leak" sanity testing, but that can come later. That effort will have to test multiple different possible causes; the manual memory load test in https://github.com/getsentry/raven-node/blob/master/test/manual/context-memory.js would not have caught this because it only exercises contexts, not each of the different autoBreadcrumb instrumentation options.

This should fix #295. As soon as this merges I'll publish a new version.

/cc @benvinegar @MaxBittker @maxcountryman

### DIFF
--- a/lib/breadcrumbs.js
+++ b/lib/breadcrumbs.js
@@ -59,43 +59,31 @@ var wrappers = {
       // possibility: capture one breadcrumb for "req sent" and one for "res recvd"
       // seems excessive but solves the problem and *is* strictly more information
       // could be useful for weird response sequencing bug scenarios
-      var self = this;
-      OrigClientRequest.call(self, options, cb);
+      OrigClientRequest.call(this, options, cb);
 
-      // We could just always grab these from self.agent, self.method, self._headers, self.path, etc
+      // We could just always reconstruct this from this.agent, this._headers, this.path, etc
       // but certain other http-instrumenting libraries (like nock, which we use for tests) fail to
       // maintain the guarantee that after calling OrigClientRequest, those fields will be populated
-      var url, method;
       if (typeof options === 'string') {
-        url = options;
-        method = 'GET';
+        this.__ravenBreadcrumbUrl = options;
       } else {
-        url = (options.protocol || '') + '//' +
+        this.__ravenBreadcrumbUrl = (options.protocol || '') + '//' +
               (options.hostname || options.host || '') +
               (options.path || '/');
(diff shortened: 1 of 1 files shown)


maxcountryman:
Nice catch! 👍

idris:
Would love to see automated tests around memory leaks, as you mentioned in the description.

This is the second time Raven introduced a memory leak into my app, with the first being #231

maxcountryman:
I would like to second @idris comment with a big plus one: Raven took down my production nodes due to this problem and it would be nice to see provisions against this happening going forward.

oliviertassinari:
@LewisJEllis Thanks for fixing it 🆗

emmenko:
Thanks for the fix, really a nasty one!!! 🎉 

FYI: this is how it looked in our grafana dashboards:
image

PR #300, opened 2017-03-24
Memory leak express patient app WIP

LewisJEllis:
Had to abuse express a bit to get this whittled down to working nicely like I wanted it to, but now it's set up to be node-clusterized so we can have a fresh process for each endpoint test so that running it all can be easy. Will finish tomorrow.

If you want to play with this for now, start up `node express-patient.js` and then either `./poke-patient.sh` or just run some of the ab commands in it.

Refs #296

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -21,6 +21,12 @@ module.exports.disableConsoleAlerts = function disableConsoleAlerts() {
 };
 
 module.exports.consoleAlert = function consoleAlert(msg) {
+  if (consoleAlerts) {
+    console.log('raven@' + ravenVersion + ' alert: ' + msg);
+  }
+};
+
+module.exports.consoleAlertOnce = function consoleAlertOnce(msg) {
   if (consoleAlerts && !(msg in consoleAlerts)) {
     consoleAlerts[msg] = true;
     console.log('raven@' + ravenVersion + ' alert: ' + msg);
@@ -135,6 +141,12 @@ function getModule(filename, base) {
   return file;
 }
 
+function parseLines(lines, frame) {
+  frame.pre_context = lines.slice(Math.max(0, frame.lineno - (LINES_OF_CONTEXT + 1)), frame.lineno - 1);
+  frame.context_line = lines[frame.lineno - 1];
+  frame.post_context = lines.slice(frame.lineno, frame.lineno + LINES_OF_CONTEXT);
+}
+
 function parseStack(err, cb) {
   var frames = [],
       cache = {};
@@ -184,7 +196,7 @@ function parseStack(err, cb) {
     }
 
     if (frame.filename in cache) {
-      parseLines(cache[frame.filename]);
+      parseLines(cache[frame.filename], frame);
       if (--callbacks === 0) cb(frames);
       return;
     }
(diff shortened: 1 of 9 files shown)


LewisJEllis:
Got the cluster stuff figured out and also played with the numbers a bunch; probably worth taking a look now for direction/sensibility @benvinegar.

Start `node manager.js` and then in another tab run `./poke-patient.sh`. First tab output looks like [this](https://gist.github.com/LewisJEllis/5cdbf06b63ed89f015297a20f4388d35).

I did this not only for breadcrumbs/contexts, but also for when we actually capture exceptions, and the notable observation was that even the simplest of exception capturing (try `ab -c 5 -n 1000 localhost:3000/capture`) seems to permanently increase the memory usage of the node process. I haven't managed to figure out why yet, but it could be cause for concern.

Left to do:
- Add nice instructions and some guidelines on how to interpret results
- Add "run this thing" to release checklist
- Figure out whats up with exception capturing memory usage (probably separate issue?)

idris:
Glad to see this!  I'm still seeing a small memory leak in my app, presumably caused by Raven.  That seems to line up with your comment "even the simplest of exception capturing ... seems to permanently increase the memory usage of the node process"

benvinegar:
Pretty cool. I played around with it for 20 minutes, but admittedly had difficulty understanding how to interpret the results at-a-glance.

> * Add nice instructions and some guidelines on how to interpret results
> * Add "run this thing" to release checklist

That'd be great.

> Figure out whats up with exception capturing memory usage (probably separate issue?)

Seems like a real problem (per @idris's comments). Let's make a new issue and see if we can get to the bottom of it next week. At least now we have a repeatable tool that can make verifying the fix is working easy.

LewisJEllis:
Wrote up a readme, have a look? It goes over the outputs/interpretations that make me believe `captureException` has a leak.

Not sure why, but on CI it seems the node-gyp build is blowing up with memwatch-next on node versions 4+; works fine locally. No super obvious issue, will investigate more; might consider just not including memwatch-next in package.json since `npm test` doesn't actually require it.

idris:
Great to see progress on this.  I should mention: The memory leak I'm seeing is also happening in a worker process with no express app running.  So it's probably not the express middleware that's the issue.
I also have all 3 breadcrumbs enabled: console, http, pg.

LewisJEllis:
@benvinegar @MaxBittker this should be good to go, including the memory leak fix

lib/utils.js
```diff
@@ -21,6 +21,12 @@ module.exports.disableConsoleAlerts = function disableConsoleAlerts() {
 };
 
 module.exports.consoleAlert = function consoleAlert(msg) {
+  if (consoleAlerts) {
+    console.log('raven@' + ravenVersion + ' alert: ' + msg);
+  }
+};
+
+module.exports.consoleAlertOnce = function consoleAlertOnce(msg) {
   if (consoleAlerts && !(msg in consoleAlerts)) {
```
benvinegar: Ah

benvinegar:
Are we merging this?

LewisJEllis:
Published 1.1.6 and 1.2.1. @idris assuming you had not done `Raven.disableConsoleAlerts()` and the small leak you were observing was the same leak I found, this should fix it. The numbers from these tests certainly show it, but as always let me know if not :)

Issue #304, opened 2017-04-10
`requestHandler` calls `next` multiple times but only logs the first error


wearhere:
**What is the current behavior?**

If Express middleware throws _multiple_ asynchronous errors, Raven's `requestHandler` will attempt to forward them all to `next`, but only the first one will reach `errorHandler`. Affects Raven 0.12.3 as well as Raven 1.2.1.

https://github.com/wearhere/raven-node-multiple-errors provides a test case. I can't tell from Express' source exactly why the second call to `next` doesn't have an effect (neither calling the downstream middleware nor causing any sort of error). It doesn't look like the second call is a pure no-op though, which is probably not good.

**What is the expected behavior?**

Every asynchronous error should be logged—I don't see why the first one to fire should be privileged.

If Express forwarded all the errors through the middleware stack I would say that's fine. But changing Express' behavior is probably wayy out of scope.

So `requestHandler` and `errorHandler` are going to have to communicate some other way. Or perhaps `requestHandler` could just log the errors itself? `requestHandler` basically calls `errorHandler` via `next` anyway, you could simplify your docs by only requiring `requestHandler`.

The documentation should also be updated to note that fall-through error handlers are _not_ guaranteed to be run on "secondary errors", unless you want to figure out how to make them run.

I would recommend not having them run though, both to avoid messing with Express and also because if the developer has a fall-through handler that terminates the request

```js
app.use((err, req, res, next) => {
  // Simplified; you wouldn't want to send down error messages in production, for instance.
  res.status(500).send(err.message);
});
```

that would not cope well being run multiple times.

LewisJEllis:
Even though your title was clear enough to me to already know what the situation/issue was, I have to give props to for the repro test case. Thanks for taking the time to provide such a nice demonstration of the issue you're describing. I agree this is a problem.

We definitely can't get rid of the `errorHandler` middleware entirely, because we want to record any errors passed down express's error handling stack; many apps follow a policy along the lines of "if we encounter an operational error that will prevent us from completing the request, just call `next(err)` and let our 500-page-error-handler deal with it", and we want to capture those errors.

But the key insight here is that we don't *have* to use express's error handling and the `errorHandler` middleware to report asynchronous exceptions captured by the domain from the `requestHandler` middleware. Currently we *do*, partially because we can and it's easy, and partially to have asynchronous errors we capture still result in the user's express app serving its 500 page or whatever else it might do.

[...]

Issue #262, opened 2017-01-26
Proper handling of unhandledRejection event


LewisJEllis:
We have a config option `captureUnhandledRejections` which is off by default but can be enabled to have Raven capture unhandled rejections automatically.

Here's what we do now:
- Capture the rejection
- Invoke the `install` callback which by default will exit the process

In the absence of Raven, unhandled rejections do not currently result in the process exiting, but enabling our `captureUnhandledRejections` option changes that.

In relation to discussion in #257 we determined that what we should do instead is:
- continue doing nothing with unhandledRejections by default
- keep the `captureUnhandledRejections` flag which adds a `process.on('unhandledRejection')` handler but does *not* result in calling `install()`'s `onFatalError` callback

This is a behavior change, but a good and necessary one. The key that really convinced me of the necessity of this change was that basically there are edge cases where an `unhandledRejection` event can be emitted but the rejection can later be handled by user application code. We don't want to incorrectly exit prematurely in that case, and this is also in line with our desire to have the exit conditions of a program not change when adding or removing Raven.

Note that unhandled rejections do result in a warning starting with Node 7, and in the future a garbage-collected unhandled rejection will result in the process exiting (see nodejs/node#6375), and this decision should make it easy for us to fall in line with that change when it happens.

LewisJEllis:
Updated this behavior in #308; will update again whenever node's default behavior changes.

Issue #257, opened 2017-01-19
Properly surfacing errors to shut down process when appropriate


LewisJEllis:
An important part of node's error handling scheme is that when an exception bubbles up to the top level uncaught (no try/catch, not in a domain, not an `error` event with a listener, etc), the process shuts down to avoid continuing to run in an unexpected and unknown state. If the program was meant to continue running under those conditions, there should have been some code somewhere to handle the exception. This philosophy is documented in [various](https://nodejs.org/api/domain.html#domain_warning_don_t_ignore_errors) [places](https://www.joyent.com/node-js/production/design/errors).

For this reason, adding raven-node to a node program should not add or remove any cases where the program shuts down due to an unhandled exception. In other words, the "when does this process crash" behavior of a program should be unaffected by the presence of raven-node, its instrumentation, handlers, context wraps, etc.

This is not currently the case and has not been the case since almost two years ago when raven-node started to catch and capture asynchronous errors in express apps via the `requestHandler` middleware (to associate them with the current request). The recent context/wrap functionality is a generalization of that. raven-node's current usage of domains results in errors that happen inside a raven-node context domain wrapper being captured, but they are then swallowed rather than making it to the top level and shutting down the process:
```javascript
wrap: function (options, func, onErr) {
  ...
  var wrapDomain = domain.create();
  wrapDomain.sentryContext = options;

  var self = this;
  if (typeof onErr !== 'function') {
    onErr = function (err) {
      // note we just capture here, we don't rethrow or otherwise propagate
      self.captureException(err);
    };
  }

  wrapDomain.on('error', onErr);
  var wrapped = wrapDomain.bind(func);
```

We can break this down into a few requirements for what we want raven-node to do/capture:
1. We want to capture any unhandled exceptions that happen, without then swallowing or ignoring them in a way that would not happen without raven present (preserve existing exit cases)
2. We want to capture some recoverable errors (like 500s in an express app), but we don't want to exit the process when it otherwise wouldn't (don't add new exit cases)
3. We want to associate any available context data like the current request, breadcrumbs, etc with captured exceptions when they happen, and avoid associating the wrong context data with the wrong exception (context collection)

Before we started to capture asynchronous errors in express apps via the requestHandler middleware, we did (1) and (2) but not (3). Now with context/wrap, we do (2) and (3) but not always (1). **We'd ultimately like to do all 3, and this issue is meant to propose/discuss the most reasonable way to get there.**

A couple additional considerations:
[...]

PR #308, opened 2017-04-27
Properly surfacing errors to preserve process exit conditions

LewisJEllis:
Two parts to this - the timing stuff, and the onFatalError stuff. Still working on the global uncaughtException part of the onFatalError stuff to make sure we capture as much as possible while avoiding recursive cases, but I tried to fill in some comments where the code is still in flux. Then I'm going to finish refactoring a bunch of tests from `patchGlobal` to `install`.

We introduce a dep on [timed-out](https://github.com/floatdrop/timed-out) which I've audited and is basically the "turn connect-timeouts and activity-timeouts into errors" logic ripped out of [request](https://github.com/request/request). This was a necessary prerequisite for us to ensure the onFatalError callback gets called no matter what happens. The timing config part might be redone to use config fields instead of dsn query param.

/cc @benvinegar

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -107,6 +107,20 @@ Those configuration options are documented below:
             sampleRate: 0.5 // send 50% of events, drop the other half
         }
 
+.. describe:: sendTimeout
+
+    The time to wait to connect to the server or receive a response when capturing an exception, in seconds.
+
+    If it takes longer, the transport considers it a failed request and will pass back a timeout error.
+
+    Defaults to 1 second.
+
+    .. code-block:: javascript
+
+        {
+            sendTimeout: 5 // wait 5 seconds before considering the capture to have failed
+        }
+
 .. describe:: dataCallback
 
     A function that allows mutation of the data payload right before being
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -113,7 +113,7 @@ Those configuration options are documented below:
 
     If it takes longer, the transport considers it a failed request and will pass back a timeout error.
 
-    Defaults to 1 second.
+    Defaults to 1 second. Make it longer if you run into timeout problems when sending exceptions to Sentry.
 
     .. code-block:: javascript
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -271,20 +271,20 @@ Global Fatal Error Handler
 --------------------------
 
 The ``install`` method sets up a global listener for uncaught exceptions, and ``context`` and ``wrap`` can catch exceptions as well.
-These are situations where Raven catches what would otherwise be a fatal process-ending exception, and we generally shouldn't keep our
-program running after such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
-so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception, we
-capture it and then call the fatal error handler.
+These are situations where Raven catches what would otherwise be a fatal process-ending exception; a process should generally not
+continue to run after uncaught exceptions and similar such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
+so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception,
+it captures the exception and then calls the fatal error handler.
 
-By default, we make the fatal error handler print the error and then exit the process. If you want to do your own clean-up,
+By default, the fatal error handler prints the error and then exits the process. If you want to do your own clean-up,
 pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as a parameter to ``install()``.
 
 The fatal error hander callback will be the last thing called before the process should shut down.
 It can do anything necessary, including asynchronous operations, to make a best effort to clean up and shut down the process, but it should
-not throw, and it absolutely must not allow the process to keep running indefinitely.
+not throw, and it absolutely must not allow the process to keep running indefinitely. This means it should probably make an explicit ``process.exit()`` call.
 
 After catching a fatal exception, Raven will first attempt to send it to Sentry before it calls the fatal exception handler.
-If sending fails, a ``sendErr`` will be passed, and otherwise the ``eventId`` will be provided. In either case, the fatal exception
+If sending fails, a ``sendErr`` error object will be passed, and otherwise the ``eventId`` will be provided. In either case, the error object 
 resulting in the shutdown is passed as the first parameter.
 
 .. code-block:: javascript
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -271,19 +271,19 @@ Global Fatal Error Handler
 --------------------------
 
 The ``install`` method sets up a global listener for uncaught exceptions, and ``context`` and ``wrap`` can catch exceptions as well.
-These are situations where Raven catches what would otherwise be a fatal process-ending exception; a process should generally not
-continue to run after uncaught exceptions and similar such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
-so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception,
-it captures the exception and then calls the fatal error handler.
+These are situations where Raven catches what would otherwise be a fatal process-ending exception. A process should generally not
+continue to run after such events occur, (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
+so Raven has a concept of a "fatal error handler". When Raven catches an otherwise-fatal exception, it will capture the exception
+(send it to Sentry) and then call the fatal error handler.
 
 By default, the fatal error handler prints the error and then exits the process. If you want to do your own clean-up,
-pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as a parameter to ``install()``.
+pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as an argument to ``install()``.
 
-The fatal error hander callback will be the last thing called before the process should shut down.
+The fatal error handler callback will be the last thing called before the process should shut down.
 It can do anything necessary, including asynchronous operations, to make a best effort to clean up and shut down the process, but it should
 not throw, and it absolutely must not allow the process to keep running indefinitely. This means it should probably make an explicit ``process.exit()`` call.
 
-After catching a fatal exception, Raven will first attempt to send it to Sentry before it calls the fatal exception handler.
+After catching a fatal exception, Raven will make a best-effort attempt to send it to Sentry before it calls the fatal exception handler.
 If sending fails, a ``sendErr`` error object will be passed, and otherwise the ``eventId`` will be provided. In either case, the error object 
 resulting in the shutdown is passed as the first parameter.
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -267,17 +267,33 @@ By default, Raven does not capture unhandled promise rejections. You can have it
     captureUnhandledRejections: true
   }).install();
 
-Global Error Handler
---------------------
+Global Fatal Error Handler
+--------------------------
 
-The ``install`` method sets up a global listener for uncaught exceptions, and possibly
-also for unhandled rejections. You generally shouldn't carry on after receiving an `uncaughtException`
-(see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
-so you can provide a callback which will be invoked **after** Raven has sent the event to Sentry:
+The ``install`` method sets up a global listener for uncaught exceptions, and ``context`` and ``wrap`` can catch exceptions as well.
+These are situations where Raven catches what would otherwise be a fatal process-ending exception, and we generally shouldn't keep our
+program running after such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
+so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception, we
+capture it and then call the fatal error handler.
+
+By default, we make the fatal error handler print the error and then exit the process. If you want to do your own clean-up,
+pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as a parameter to ``install()``.
+
+The fatal error hander callback will be the last thing called before the process should shut down.
+It can do anything necessary, including asynchronous operations, to make a best effort to clean up and shut down the process, but it should
+not throw, and it absolutely must not allow the process to keep running indefinitely.
+
+After catching a fatal exception, Raven will first attempt to send it to Sentry before it calls the fatal exception handler.
+If sending fails, a ``sendErr`` will be passed, and otherwise the ``eventId`` will be provided. In either case, the fatal exception
+resulting in the shutdown is passed as the first parameter.
 
 .. code-block:: javascript
 
-    Raven.install(function() {
+    Raven.install(function (err, sendErr, eventId) {
+      if (!sendErr) {
+        console.log('Successfully sent fatal error with eventId ' + eventId + ' to Sentry:');
+        console.error(err.stack);
+      }
       console.log('This is thy sheath; there rust, and let me die.');
       process.exit(1);
     });
--- a/lib/client.js
+++ b/lib/client.js
@@ -106,6 +106,14 @@ extend(Raven.prototype, {
     return this;
   },
 
+  setOnFatalErrorCallback: function (cb) {
+    this.onFatalError = cb;
+    // ensure that we use this handler
+    // - after uncaughtExceptions
+    // - after domain context captures
+    registerExceptionHandler(this, cb);
+  },
+
   install: function install(opts, cb) {
     if (this.installed) return this;
 
@@ -113,10 +121,13 @@ extend(Raven.prototype, {
       cb = opts;
     }
 
-    registerExceptionHandler(this, cb);
-    if (this.captureUnhandledRejections) {
-      registerRejectionHandler(this, cb);
+    if (typeof cb !== 'function') {
+      cb = function (err) {
+        console.error(err);
+        process.exit(1);
+      };
     }
+    this.setOnFatalErrorCallback(cb);
 
     for (var key in this.autoBreadcrumbs) {
       if (this.autoBreadcrumbs.hasOwnProperty(key)) {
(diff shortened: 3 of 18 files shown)


lib/utils.js
```diff
@@ -68,6 +69,9 @@ module.exports.parseDSN = function parseDSN(dsn) {
           host: parsed.host.split(':')[0]
         };
 
+    var timeout = qs.parse(parsed.query).timeout;
```
benvinegar: Wait, timeout is part of the DSN?

lib/client.js
```diff
@@ -488,40 +497,58 @@ nodeUtil.inherits(Client, Raven);
...
-      if (called) {
-        client.removeListener('logged', onLogged);
-        client.removeListener('error', onError);
-        return cb(false, err);
-      }
+    // if (cb) { // bind event listeners only if a callback was supplied
+    //   var onLogged = function onLogged() {
+    //     called = false;
+    //     cb(true, err);
+    //   };
```
benvinegar: It would be easier if this code was just removed re: reviewing.

lib/client.js
```diff
@@ -488,40 +497,58 @@ nodeUtil.inherits(Client, Raven);
...
+    if (!calledUncaughtHandler) {
+      // this might be overly careful here - need to avoid getting messed up by recursion
+      calledUncaughtHandler = true;
+      var eventId = client.captureException(err, function (sendErr, eventId) {
+        try {
+          client.onFatalError(err);
+        } catch (e) {
+          // no-op just avoid inf loop...
+        }
+        // consider signature like client.onFatalError(err, sendErr, eventId);
```
benvinegar: This does exist as of this PR though, right? (`setOnFatalErrorCallback`)

lib/utils.js
LewisJEllis: This ties into the "should we configure raven clients via DSN querystring params or via config fields and how should we standardize that stuff" discussion; I originally followed the lead of [raven-java](https://docs.sentry.io/clients/java/config/) and [raven-python](https://github.com/getsentry/raven-python/blob/master/docs/transports.rst) which both do it via DSN, but we don't really have any precedent of doing DSN-based config in raven-node. Open for discussion.

lib/client.js
LewisJEllis: setOnFatalErrorCallback exists, yes, and it stores the callback in the `onFatalError` property - this comment is regarding the question "what should we pass to onFatalError when we invoke it"

lib/client.js
LewisJEllis: Sure, but WIP etc; I'll clean up a bit more in the future.

LewisJEllis:
Yes, this will definitely be semver major due to change in when-will-program-exit behavior alone. uncaughtException recursion concern is mostly figured out, test and doc changes in progress.

LewisJEllis:
I ended up shelling out to separate processes for the exit condition tests, but that turned out to be much more feasible than what I was trying before. It's a little messy in coupling the separate test script with the mocha assertions or duplicating instantiation/nock/etc in the test scripts, but I think it works well enough.

I also finally figured out the right thing to be doing for recursive cases, errors thrown during capture attempts, etc, and how to reconcile that between both uncaughtException and domain cases. This also resulted in a change to middleware behavior described in #304. 

Left to do:
- change timeout to be a config field instead of DSN querystring
- write the docs

LewisJEllis:
This is all ready to review; the updated docs should give a brief overview of new behavior, and #257 should provide remaining background.

/cc @benvinegar @MaxBittker

docs/usage.rst
```diff
@@ -267,17 +267,33 @@ By default, Raven does not capture unhandled promise rejections. You can have it
...
 
-The ``install`` method sets up a global listener for uncaught exceptions, and possibly
-also for unhandled rejections. You generally shouldn't carry on after receiving an `uncaughtException`
-(see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
-so you can provide a callback which will be invoked **after** Raven has sent the event to Sentry:
+The ``install`` method sets up a global listener for uncaught exceptions, and ``context`` and ``wrap`` can catch exceptions as well.
+These are situations where Raven catches what would otherwise be a fatal process-ending exception, and we generally shouldn't keep our
+program running after such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
+so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception, we
+capture it and then call the fatal error handler.
```
MaxBittker: this we is fine IMO because it's referring to sentry authors. but consider changing to "raven will capture it"

docs/usage.rst
```diff
@@ -267,17 +267,33 @@ By default, Raven does not capture unhandled promise rejections. You can have it
...
+
+After catching a fatal exception, Raven will first attempt to send it to Sentry before it calls the fatal exception handler.
+If sending fails, a ``sendErr`` will be passed, and otherwise the ``eventId`` will be provided. In either case, the fatal exception
+resulting in the shutdown is passed as the first parameter.
 
 .. code-block:: javascript
 
-    Raven.install(function() {
+    Raven.install(function (err, sendErr, eventId) {
+      if (!sendErr) {
```
MaxBittker: sendErr is confusing, because it's a error in sending the error. maybe name this something longer and more explicit? like "sendFailed"?

docs/config.rst
```diff
@@ -107,6 +107,20 @@ Those configuration options are documented below:
             sampleRate: 0.5 // send 50% of events, drop the other half
         }
 
+.. describe:: sendTimeout
+
+    The time to wait to connect to the server or receive a response when capturing an exception, in seconds.
+
+    If it takes longer, the transport considers it a failed request and will pass back a timeout error.
+
+    Defaults to 1 second.
```
MaxBittker: I wouldn't mind a a sentence here about why you might want to set this longer or shorter but ndb

docs/usage.rst
LewisJEllis: `sendFailed` sounds like a boolean, and it's a common naming convention for any error object to be named `somethingErr`; could call it `sendingErr` but the present progressive has other undertones/implications that I don't really like either.

LewisJEllis:
Updated unhandledRejection behavior as per #262. This is now essentially just:
```javascript
process.on('unhandledRejection', function (reason) {
  Raven.captureException(reason);
});
```
but the reasons for keeping the config field instead of just telling people in the docs to do their own handler like above are:
  1. backwards compat/it already exists
  2. the behavior of unhandled rejections is likely to change in the future (see nodejs/node#12734) and we want to be able to update our behavior to stay in line as mentioned at the end of #262.

docs/usage.rst
MaxBittker: transmissionError?

lib/client.js
```diff
@@ -45,6 +45,7 @@ extend(Raven.prototype, {
     this.name = options.name || process.env.SENTRY_NAME || require('os').hostname();
     this.root = options.root || process.cwd();
     this.transport = options.transport || transports[this.dsn.protocol];
+    this.sendTimeout = options.sendTimeout || 1;
```
MaxBittker: would we ever be more defensive about types here? I could imagine something that checks for a number less than 1000 but we don't do that anywhere else in options

lib/client.js
```diff
@@ -99,23 +100,34 @@ extend(Raven.prototype, {
...
     this.on('error', function (err) {
       utils.consoleAlert('failed to send exception to sentry: ' + err.message);
     });
 
     return this;
   },
 
-  install: function install(opts, cb) {
+  install: function install(cb) {
     if (this.installed) return this;
```
MaxBittker: what happened to opts, was that just never used?

lib/client.js
LewisJEllis: Yea, there are a lot of situations where we're a little flexible/not-defensive but where it's fine as long as users aren't idiots. Since thus far there haven't been too many cases of that, I stick to the generally existing level of defensiveness; if we want to dial it up, there are lots of spots to add typeofs and bad-input-early-returns.

lib/client.js
LewisJEllis: It was initially used for passing autobreadcrumb and captureUnhandledRejections configs; we decided to move that to part of `config()`, so it was unused and I just hadn't removed it.

lib/client.js
```diff
@@ -143,6 +155,30 @@ extend(Raven.prototype, {
...
 
+  makeErrorHandler: function () {
+    var self = this;
+    var calledUncaughtHandler = false;
+    var calledFatalError = false;
+    var firstError;
+    return function (err) {
+      if (!calledUncaughtHandler) {
+        firstError = err;
+        calledUncaughtHandler = true;
```
MaxBittker: would you consider getting rid of calledUncaughtHandler and using firstError being truthy instead? could be less to juggle here.

LewisJEllis: I did consider this, but it's possible to literally `throw false` or `throw undefined`:
```javascript
process.on('uncaughtException', function (err) {
  console.log('uncaught: ', typeof err);
});
throw undefined;
```
outputs:
```
uncaught: undefined
```
😅

MaxBittker: `firstError = {value: err};` 😅

lib/client.js
```diff
@@ -143,6 +155,30 @@ extend(Raven.prototype, {
...
+        self.captureException(err, function (sendErr, eventId) {
+          calledFatalError = true;
+          self.onFatalError(err, sendErr, eventId);
+        });
+      } else if (!calledFatalError) {
+        self.onFatalError(firstError, err);
+      } else {
+        // we hit an error *after* calling onFatalError - pretty fucked
```
MaxBittker: it's prrrrrobably better not to say 'fucked' -- it won't be in the minified code but I could imagine some kind of automatic censorship causing a problem or something. idk

lib/client.js
LewisJEllis: 🙄 

booleans are cheaper than objects 💵

lib/client.js
LewisJEllis: Fair point. I ran the numbers on the codebase's current 🚽-mouth stature:

| Word | # |
| ---- | - |
| Crap | 6 |
| Fuck | 3 |
| Shit | 1 |

And for commit messages:
| Word | # |
| ---- | - |
| Crap | 0 |
| Fuck | 1 |
| Shit | 1 |

We can put things through the 🛁  in a later commit.

lib/client.js
```diff
@@ -285,14 +321,7 @@ extend(Raven.prototype, {
...
 
-    var self = this;
-    if (typeof onErr !== 'function') {
-      onErr = function (err) {
-        self.captureException(err);
-      };
-    }
-
-    wrapDomain.on('error', onErr);
+    wrapDomain.on('error', this.makeErrorHandler());
```
MaxBittker: what are the implications of having multiple makeErrorHandler() closures floating around? aren't they there to do some sort of de-duplication?

LewisJEllis: meant to encapsulate the `firstError` and `calledWhatever` flags; we have a separate fatal error catcher/handler for each domain and then one for global uncaught exceptions. since domains are sort of heavy anyway, I don't consider it a big cost. it miiight be possible to combine all the domain ones to one set of flags, but I feel like there'd be some potential resulting contrived probably-never-happen-but-might-mess-us-up-if-it-ever-did cases that I'd rather not go thinking about if we can avoid it for relatively little cost.

test/exit/capture_callback.js
```diff
@@ -0,0 +1,23 @@
...
+
+Raven.config(dsn).install(function (err, sendErr, eventId) {
+  scope.done();
+  assert.equal(sendErr, null);
+  assert.ok(err instanceof Error);
+  assert.equal(err.message, 'derp');
+  console.log('exit test assertions complete');
+  process.exit(20);
+});
+
```
MaxBittker: should you have a` process.on('exit',()=>{scope.done()}`?

LewisJEllis: I don't see any need for it, what's the case you're thinking it would help check against? We already check for `scope.done()` at the beginning of the onFatalError callback, and the mocha test case checks that this test exits with code 20, so the only exit path that will pass there has to go through the `scope.done()` here.

MaxBittker: ah missed that aspect of the mocha test, 👍

lib/client.js
MaxBittker: no this is good now that I understand

lib/client.js
```diff
@@ -99,23 +100,34 @@ extend(Raven.prototype, {
       globalContext.extra = options.extra;
     }
 
+    this.onFatalError = this.defaultOnFatalError = function (err, sendErr, eventId) {
+      console.error(err.stack);
```
MaxBittker: `cannot read property 'stack' of undefined` <- this could stop us from exiting, right? whats the guarantees here around err?

MaxBittker: still curious what you think about this though.

LewisJEllis: good catch, I was just doing `console.error(err)` but changed to `err.stack` for some goofy domain cases, will make this more defensive

MaxBittker:
This review is pending the extra commits to change the successive error behavior.

lib/client.js
```diff
@@ -101,9 +101,10 @@ extend(Raven.prototype, {
     }
 
     this.onFatalError = this.defaultOnFatalError = function (err, sendErr, eventId) {
-      console.error(err.stack);
+      console.error(err && err.stack ? err.stack : err);
```
MaxBittker: out of curiosity, can you build this only with short circuiting and no ternary? :P

MaxBittker: `err && err.stack || err`  I think? probably don't actually do this...

lib/client.js
```diff
@@ -157,22 +158,50 @@ extend(Raven.prototype, {
...
+            // it was probably case 1, let's treat err as the sendErr and call onFatalError
+            calledFatalError = true;
+            self.onFatalError(firstError, err);
+          } else {
+            // it was probably case 2, our first error finished capturing while we waited, cool, do nothing
+          }
+        }, (self.sendTimeout + 1) * 1000); // capturing could take at least sendTimeout to fail, plus an arbitrary second for how long it takes to collect surrounding source etc
+      } else if (calledFatalError) {
+        // we hit an error *after* calling onFatalError - pretty boned at this point, just shut it down
         self.consoleAlert('uncaught exception after calling fatal error shutdown callback - this is bad! forcing shutdown');
```
MaxBittker: could you add a test case for this path? in the name of coverage?

LewisJEllis: Yea, sure, why not

lib/client.js
LewisJEllis: maybe, but for same "people can throw dumb stuff" reason as before I'd rather just leave it explicit like this

lib/client.js
LewisJEllis: @MaxBittker done, good call, had to fix a thing

lib/client.js
MaxBittker: i was just curious, yours it better

lib/client.js
MaxBittker: lol i missed these tables, love it

docs/usage.rst
```diff
@@ -271,19 +271,19 @@ Global Fatal Error Handler
 --------------------------
 
 The ``install`` method sets up a global listener for uncaught exceptions, and ``context`` and ``wrap`` can catch exceptions as well.
-These are situations where Raven catches what would otherwise be a fatal process-ending exception; a process should generally not
-continue to run after uncaught exceptions and similar such events (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
-so Raven has a concept of a "fatal error handler". When Raven catches what would otherwise be a fatal process-ending exception,
-it captures the exception and then calls the fatal error handler.
+These are situations where Raven catches what would otherwise be a fatal process-ending exception. A process should generally not
+continue to run after such events occur, (see `Node docs <http://nodejs.org/api/process.html#process_event_uncaughtexception>`_),
+so Raven has a concept of a "fatal error handler". When Raven catches an otherwise-fatal exception, it will capture the exception
```
MaxBittker: I think we should make it clear that we will capture *and send* the error, because someone might think they need to send it themselves in the fatal error handler

docs/usage.rst
```diff
@@ -271,19 +271,19 @@ Global Fatal Error Handler
...
-pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as a parameter to ``install()``.
+pre-exit logging, or other shutdown procedures, you can provide your own fatal error handler as an argument to ``install()``.
 
-The fatal error hander callback will be the last thing called before the process should shut down.
+The fatal error handler callback will be the last thing called before the process should shut down.
 It can do anything necessary, including asynchronous operations, to make a best effort to clean up and shut down the process, but it should
 not throw, and it absolutely must not allow the process to keep running indefinitely. This means it should probably make an explicit ``process.exit()`` call.
 
-After catching a fatal exception, Raven will first attempt to send it to Sentry before it calls the fatal exception handler.
+After catching a fatal exception, Raven will make a best-effort attempt to send it to Sentry before it calls the fatal exception handler.
```
MaxBittker: you do say it down here but maybe this could be condensed

docs/usage.rst
LewisJEllis: yea so we generally the term "capture" means "send" but ill see if i can make that more explicit

lib/client.js
```diff
@@ -143,6 +156,58 @@ extend(Raven.prototype, {
...
+        caughtFirstError = true;
+        self.captureException(err, function (sendErr, eventId) {
+          if (!calledFatalError) {
+            calledFatalError = true;
+            self.onFatalError(err, sendErr, eventId);
+          }
+        });
+      } else if (calledFatalError) {
+        // we hit an error *after* calling onFatalError - pretty boned at this point, just shut it down
+        utils.consoleAlert('uncaught exception after calling fatal error shutdown callback - this is bad! forcing shutdown');
```
MaxBittker: what if on fatal error is async? couldn't this stop it from running?

MaxBittker: i know this sucks but it seems like this should possibly have another timeout

LewisJEllis: Yea, if there's another independent error after onFatalError starts running we're just gonna shut down, but I'm not sure if we should, or even *can* reliably, worry about ensuring the completion of onFatalError - there's no way for us to really know how long onFatalError should take, whereas we do have some rough idea of how long capturing the first error should take. I'm much more worried about ensuring that we successfully capture the original first error and that the process shuts down.

It could make sense to expose some sort of user option for deciding where we should sit on the bail-at-first-trouble vs wait-it-out-and-see spectrum; right now as long as fatalError has been called, we bail if any further bad stuff happens, mostly in order to ensure the process exits and doesn't hang. I'm not sure what that option would look like or what the use case for it would be, so I kind of want to defer it; anything being done in the onFatalError callback is a best-effort no-guarantees sort of thing anyway.

MaxBittker: I think we should talk about this more next week but it's fine for now. nice PR 👍

benvinegar:
gj team

PR #309, opened 2017-05-05
Don't include domain prop as extra property on Error objects

LewisJEllis:
An error caught by a domain handler (from `context()` or `requestHandler`) will have a `domain` property. Domains can have references to all kinds of stuff - Express layers, server sockets, who knows what, just depends on what functions ran in the domain - and we don't really get any value from including it as an extra property on error objects like we currently do.

It's potentially confusing, could include random (maybe-sensitive) server state, and is a bunch (sometimes a *whole* bunch) of data for little value.

Previously:

After:

/cc @MaxBittker @benvinegar

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -31,7 +31,7 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
     var extraErrorProps;
     for (var key in err) {
       if (err.hasOwnProperty(key)) {
-        if (key !== 'name' && key !== 'message' && key !== 'stack') {
+        if (key !== 'name' && key !== 'message' && key !== 'stack' && key !== 'domain') {
           extraErrorProps = extraErrorProps || {};
           extraErrorProps[key] = err[key];
         }


MaxBittker:
we still send the user / context info that we put in the domain in the first place?

lib/parsers.js
```diff
@@ -31,7 +31,7 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
     var extraErrorProps;
     for (var key in err) {
       if (err.hasOwnProperty(key)) {
-        if (key !== 'name' && key !== 'message' && key !== 'stack') {
+        if (key !== 'name' && key !== 'message' && key !== 'stack' && key !== 'domain') {
```
MaxBittker: maybe consider some newlines in this expression

LewisJEllis:
> we still send the user / context info that we put in the domain in the first place?

Yes, of course; not all errors with a context are caught by a domain handler, so we couldn't be relying on this anyway. The domain context handling is entirely separate. This part of parseError is just going "hey, this Error has extra properties on it, maybe those are useful and we'd like to see them".

lib/parsers.js
LewisJEllis: I personally dislike newlines in the middle of an if conditional because of how the indentation ends up looking, so I avoid them as much as possible, but the alternative I would consider is:
```javascript
if (['name', 'message', 'stack', 'domain'].indexOf(key) === -1) {
```
Arguable whether this is more readable, I could go either way.

MaxBittker:
``` if(key !== 'name' &&
    key !== 'message' &&
    key !== 'stack' &&
    key !== 'domain') {
 
 }``` up to you, but indexOf is prob worse

PR #92, opened 2014-09-12
Automatically parse request in captureMessage/captureError (#82)

rcoup:
As discussed in #82, automatically find a request object passed in kwargs to captureMessage()/captureError() and parse it.

``` javascript
client.captureMessage("hi", {request:req});
client.captureError(err, {request:req});
```

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-node/92)

PR #310, opened 2017-05-05
Parse req from context, and requestHandler sets req on context

LewisJEllis:
This basically implements a slightly more thorough version of what I described in [this comment](https://github.com/getsentry/raven-node/issues/82#issuecomment-281935866) in #82, and should also address what #92 and #93 were trying to solve.

Basically you can now do:
```javascript
Raven.setContext({ req: req });
```
and then we'll parse `req` as a node/express/koa request object the same way the `errorHandler` middleware does. We'll do the same for a req passed via kwargs to a captureException call:
```javascript
Raven.captureException(myError, { req: req });
```
and the `requestHandler` middleware now automatically adds `req` to the context (i.e. does this for you).

/cc @MaxBittker @benvinegar

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -228,6 +228,13 @@ extend(Raven.prototype, {
       values: domainContext.breadcrumbs || this._globalContext.breadcrumbs || []
     };
 
+    kwargs.request = extend({}, this._globalContext.request, domainContext.request, kwargs.request);
+    if (Object.keys(kwargs.request).length === 0) {
+      var req = extend({}, this._globalContext.req, domainContext.req, kwargs.req);
+      var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
+      kwargs = extend(kwargs, parsers.parseRequest(req, parseUser));
+    }
+
     kwargs.modules = utils.getModules();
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -228,6 +228,13 @@ extend(Raven.prototype, {
       values: domainContext.breadcrumbs || this._globalContext.breadcrumbs || []
     };
 
+    kwargs.request = extend({}, this._globalContext.request, domainContext.request, kwargs.request);
+    if (Object.keys(kwargs.request).length === 0) {
+      var req = extend({}, this._globalContext.req, domainContext.req, kwargs.req);
+      var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
+      kwargs = extend(kwargs, parsers.parseRequest(req, parseUser));
```
MaxBittker: extend mutates kwargs without the assignment but if you find this clearer its up to you

MaxBittker: I think you need some comments here about req vs request

MaxBittker: because i'm slightly confused even after reading the comment on the issue

LewisJEllis: Good call, added some commentary.

PR #314, opened 2017-05-11
Version 2.0.0

LewisJEllis:
Accepting feedback on the release notes. Will publish first thing tomorrow.

/cc @MaxBittker @benvinegar

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,18 @@
+# 2.0.0 - 5/10/2017
+- Properly surface errors to preserve process exit conditions [See #308, #257]
+  - Node processes with raven will now exit in exactly the same situations as if raven were not present
+    - Previously, there were failure scenarios where raven would erroneously cause the process to continue to run when it should have shut down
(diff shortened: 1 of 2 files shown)


benvinegar:
Based on our previous discussion, this seems right, but I think I'll leave to @MaxBittker to approve.

History.md
```diff
@@ -1,3 +1,16 @@
+# 2.0.0 - 5/10/2017
+- Properly surface errors to preserve process exit conditions [See #308, #257]
+  - Node processes with raven will now exit in *all* the same situations as if raven were not present
```
MaxBittker: maybe don't emphasize all here when it's not a hard hard guarantee

MaxBittker: just say "the same situations" imo

LewisJEllis:
> which points someone needs to be aware of for upgrading?
Mostly just the one point about shutdown conditions; added a bold callout.

Added the postgres autobreadcrumb removal to this.

PR #315, opened 2017-05-11
Remove pg instrumentation

LewisJEllis:
Since we're publishing 2.0.0 we can do this for temporary relief for #254, as discussed privately.

/cc @MaxBittker @benvinegar


LewisJEllis:
Sure, updated.

Issue #316, opened 2017-05-12
Error raven.lib:client in Raven.captureException


gnumoksha:
I am getting a lot of "Error raven.lib:client in Raven.captureException" on my sentry dashboard that is monitoring my [parse-servers](https://github.com/parse-community/parse-server). I think [this](https://github.com/getsentry/raven-node/issues/187) issue is about it, but is not conclusive.

The message is:
```
Error: [object Object]
```

And the exception is:
```
Error: [object Object]
  File "/var/app/current/node_modules/raven/lib/client.js", line 242, col 13, in Raven.captureException
    err = new Error(err);
  File "/var/app/current/node_modules/raven/lib/client.js", line 420, col 26, in Layer.<anonymous>
    var eventId = self.captureException(err, kwargs);
  File "/var/app/current/node_modules/newrelic/lib/instrumentation/express.js", line 501, col 21, in Layer.wrappedHandle
    return handle.apply(this, args)
  File "<anonymous>", line 4, col 28, in eval (eval at wrapHandle (/var/app/current/node_modules/newrelic/lib/instrumentation/express.js:1:0),
  File "/var/app/current/node_modules/express/lib/router/layer.js", line 58, col 5, in Layer.handle_error
    fn(error, req, res, next);
  File "/var/app/current/node_modules/newrelic/lib/instrumentation/express.js", line 228, col 32, in Layer.wrappedLayerHandleError [as handle_error]
    return handleError.apply(this, arguments)
  File "/var/app/current/node_modules/express/lib/router/index.js", line 300, col 13, in trim_prefix
    layer.handle_error(layerError, req, res, next);
  File "/var/app/current/node_modules/express/lib/router/index.js", line 270, col 7, in null.<anonymous>
    trim_prefix(layer, layerError, layerPath, path);
  File "/var/app/current/node_modules/express/lib/router/index.js", line 321, col 12, in Function.proto.process_params
    return done();
  File "/var/app/current/node_modules/newrelic/lib/instrumentation/express.js", line 182, col 31, in Function.cls_wrapProcessParams [as process_params]
    return process_params.apply(this, arguments)
  File "/var/app/current/node_modules/express/lib/router/index.js", line 261, col 10, in next
    self.process_params(layer, paramcalled, req, res, function (err) {
  File "/var/app/current/node_modules/newrelic/lib/transaction/tracer/index.js", line 184, col 28, in wrapped
    var result = handler.apply(this, arguments)
[...]

PR #317, opened 2017-05-14
Handles a user-defined error that was not inherited from Error.prototype (#316)

gnumoksha:


gnumoksha:
Hi. Only the commit 3df90fccbc69e8434cd4bb10aea6cd3bae1ecd6c needs to be merged. As I need the fix, I will be using my fork until this commit has some feedback.

kamilogorek:
Unfortunately this is not correct `Error` implementation, as it completely loses `stack` property of an error. This change will break desired behaviour when someone throws an object literal (which sometimes happen), by stringifying it before passing to an `Error` constructor (line 328).

If we'd need a feature like this, we'd need to make sure that every error has a specific interface (code, message, stack) and follows specific convention. More on this here: https://gist.github.com/slavafomin/b164e3e710a6fc9352c934b9073e7216

Closing for now, but please feel free to reopen if you'd want to continue the discussion.

PR #320, opened 2017-05-16
Parsestack refactor + tests to fix memory leak

### LINKED ISSUE #319: Out of memory in parseStack.
My node app starts with ~250MB ram usage.

When an error occur, the ram usage spikes over 2GB and the node process crashes because it runs out of the default RAM limit (~1.75GB).

Upon investigation, it is the Raven code that consumes all that ram.

The utils.parseStack() implements a simple cache to avoid parsing the same javascript file multiple times if there is more than one stack frame on the same filename for the error.

It performs an fs.readFile() which is an async call, and adds the result to the cache on success.

My app is a webpack bundle on the server-side, and it results in an out of memory error because the javascript bundle is about 9mb and is being parsed over 20 times concurrently.

It seems like the cache is not working, the file should be parsed only once. 

What happens is that all the frames will generate a cache-miss, and then populate the same cache key on success... rendering the cache useless.

Relevant piece of code is here:
https://github.com/getsentry/raven-node/blob/master/lib/utils.js#L204-L213

You can see the cache key is set only on success.

A possible fix would be to split the code into 2 parts: First, loop over the frames and load all files, fully populating the cache. 
Wait for all cache entries to be loaded, then perform the parseLines() on each stack frame.

LewisJEllis:
Fixes #319, based on the insight in #318; thanks to @asuka999 for the original insight + PR and @cveilleux for the memory leak report.

- refactors `parseStack`into a more serial-procedural style with a `readSourceFiles` helper
- adds a unit test to check that readFile is only called once per file
- adds a manual memory test for capturing an error thrown from 1000-deep callstack in a 5MB file 
  - without this fix, parsing that stack and reading it multiple times concurrently would run the node process out of memory

The parseStack code in general needed some overhauling, so I went ahead and did that in addition to just fixing the concurrency control flow. It should hopefully be more robust and easier to follow now.

/cc @MaxBittker @benvinegar

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,7 +6,6 @@ var transports = require('./transports');
 var path = require('path');
 var lsmod = require('lsmod');
 var stacktrace = require('stack-trace');
-var Promise = require('promise/domains');
 
 var ravenVersion = require('../package.json').version;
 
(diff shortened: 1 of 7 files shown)


LewisJEllis:
One notable point I forgot to mention before is that the read-source-files-cache is on a per-parseStack basis, so the cached files are not reused between multiple parseStack calls. This is the same as it was before and my statement in #318 that
> we won't read a file again if a prior parseStack call already read it

was incorrect. I believe this is the correct default behavior to ensure that in a long-running process we won't accumulate a large cache which ends up consuming a significant amount of memory, but it makes for higher disk read load in a typical case; I think a future option for whether to keep the cache around between calls could be useful for users who are more concerned about disk read load than about memory usage.

lib/utils.js
```diff
@@ -142,84 +141,77 @@ function getModule(filename, base) {
...
+  return readSourceFiles(Object.keys(filesToRead), function (sourceFiles) {
+    frames.forEach(function (frame) {
+      if (frame.filename && sourceFiles[frame.filename]) {
+        var lines = sourceFiles[frame.filename];
+        try {
+          frame.pre_context = lines.slice(Math.max(0, frame.lineno - (LINES_OF_CONTEXT + 1)), frame.lineno - 1);
+          frame.context_line = lines[frame.lineno - 1];
+          frame.post_context = lines.slice(frame.lineno, frame.lineno + LINES_OF_CONTEXT);
+        } catch (e) {
+          // anomaly, being defensive in case
```
MaxBittker: do you want to set the properties to be empty strings here?

test/raven.utils.js
```diff
@@ -168,6 +168,44 @@ describe('raven.utils', function () {
...
+        var frame1 = frames.pop();
+        var frame2 = frames.pop();
+        frame1.context_line.trim().should.endWith("throw new Error('error');");
+        frame2.context_line.trim().should.endWith('nestedThrow();');
+        frame1.filename.should.equal(frame2.filename);
+
+        var uniqueFilesRead = filesRead.filter(function (filename, idx, arr) {
+          return arr.indexOf(filename) === idx;
+        });
+        filesRead.length.should.equal(uniqueFilesRead.length);
```
MaxBittker: 👍

lib/utils.js
```diff
@@ -142,84 +141,77 @@ function getModule(filename, base) {
...
 
-  var callbacks = stack.length;
-
-  // Sentry requires the stack trace to be from oldest to newest
+  // Sentry expects the stack trace to be oldest -> newest, v8 provides newest -> oldest
   stack.reverse();
 
-  return stack.forEach(function (line, index) {
+  var frames = [];
+  var filesToRead = {};
```
MaxBittker: could new Object(null) here if you wanted idk

lib/utils.js
LewisJEllis: nah, i think we'd rather not send them to the server at all; not sure what behavior diff would be but not sending them at all is what we do for native frames so i'll stick to that

MaxBittker: agreed

lib/utils.js
```diff
@@ -142,84 +141,77 @@ function getModule(filename, base) {
...
 
-function readFilePromise(filename) {
-  return new Promise(function (resolve, reject) {
-    fs.readFile(filename, function (err, file) {
-      return err ? reject(err) : resolve(file);
+  var sourceFiles = {};
+  var numFilesToRead = filenames.length;
+  return filenames.forEach(function (filename) {
+    fs.readFile(filename, function (readErr, file) {
+      if (!readErr) sourceFiles[filename] = file.toString().split('\n');
```
MaxBittker: should you wrap this line in a try catch, in case toString throws due to a weird file? to ensure that the callback always fires.

lib/utils.js
LewisJEllis: :+1:

lib/utils.js
LewisJEllis: `if (!readErr)` guarantees that `file` is a [Buffer](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end) according to [fs.readFile docs](https://nodejs.org/api/fs.html#fs_fs_readfile_file_options_callback), so I'm not concerned

MaxBittker: ok 👍

PR #322, opened 2017-05-22
Instrument require() for autobreadcrumbs

LewisJEllis:
This is a bit behind where I am locally, will push an update soon, but here's the PR to start tracking.


LewisJEllis:
This is now pretty much fully working. I had a bunch of subtle module caching and double-instrumentation stuff going on in the tests because my first pass at `uninstall()` was slightly insufficient, but got that worked out now.

Still want to investigate a couple things:
- http breadcrumbs are sometimes missing `req.method`, unsure if related to these changes or a separate thing
- option of hooking `Module._load` instead of `Module.prototype.require`: pros and cons, any reason to prefer one or the other?
- observation: when node-core's `https` module does its own `require('http')`, it won't hit our hooked `require` to instrument `http`, but if we've previously instrumented `http` it *will* get our instrumented version, hence the proactive `require('http')` in `instrumentor.instrument`;  I want to actually *understand why* this is.

/cc @MaxBittker @benvinegar

benvinegar:
> This is now pretty much fully working.

> Still want to investigate a couple things:

Is this still WIP? Should we be reviewing?

LewisJEllis:
Worked through investigation points 2 and 3 and determined that current state is acceptable. Identified a duplication issue of some sort in point 1, tracking it down and then I'll remove WIP tag and request review.

LewisJEllis:
Tracked down duplication. Also *actually* figured out the explanation behind point 3, but opted to leave it as is for to avoid a significant mostly-unnecessary refactor; explained in a comment above where we proactively `require('http')`.

Ready for review; primary piece to review is instrumentor.js, how that's used from client.js and the coupling between instrumentor and Raven, and the interface between instrumentor and each of the three "instrument this module" files.

lib/instrumentation/instrumentor.js
```diff
@@ -0,0 +1,77 @@
...
+  console: true
+};
+
+var defaultConfig = {
+  console: false,
+  http: false,
+  pg: false
+};
+
+function fill(obj, name, replacement, track) {
```
benvinegar: Can we not put this in `../utils`? It's duplicated 3 times in this patch.

lib/instrumentation/instrumentor.js
```diff
@@ -0,0 +1,77 @@
...
+    config = utils.extend({}, defaultConfig, config);
+  }
+
+  Raven.instrumentedOriginals = [];
+  Raven.instrumentedModules = [];
+
+  var Module = require('module');
+  fill(Module, '_load', function (origLoad) {
+    return function (moduleId, parent, isMain) {
+      var origModule = origLoad.apply(this, arguments);
```
benvinegar: Worth doing that `Array` args copy / perf trick here?

lib/client.js
```diff
@@ -51,26 +52,9 @@ extend(Raven.prototype, {
...
-    // this procedure will ensure that this.autoBreadcrumbs is an object populated
-    // with keys -> bools reflecting actual status of all breadcrumb types
-    var autoBreadcrumbDefaults = {
-      console: false,
-      http: false,
-    };
+    this.autoBreadcrumbs = options.autoBreadcrumbs || false;
     // default to 30, don't allow higher than 100
     this.maxBreadcrumbs = Math.max(0, Math.min(options.maxBreadcrumbs || 30, 100));
-    this.autoBreadcrumbs = extend({}, autoBreadcrumbDefaults);
```
benvinegar: I thought we had a test for this logic. Is it not affected by this?

lib/instrumentation/instrumentor.js
LewisJEllis: Maybe, but there are a few other spots where we're doing `something.apply(this, arguments)` in instrumentation for console and http (and fwiw raven-js also has spots where it doesn't do the copy opt...), so I'd rather go all-or-nothing on it.

lib/client.js
LewisJEllis: For which part in particular? I basically moved the "figure out what our autobreadcrumb to turn on" bit into instrumentor. I don't think anything really changed behavior-wise, just where it lives, so now client has to care a lot less and just has to pass its config on to the instrumentor. It actually ended up simplifying things quite a bit from a combination of two or three small differences.

lib/instrumentation/instrumentor.js
LewisJEllis: Yea, sure, I had a mental note on this but forgot about it. This was another of my motivating factors for a shared js utils, but that's a separate issue.

PR #324, opened 2017-05-24
Don't parse empty `req` properties

### LINKED ISSUE #321: kwargs.request is being set when there is no request
When using v2.0.1 on a non-web application (i.e. a worker process), the client is setting `kwargs.request` with garbage information which then clutters up the issue page when viewing the exception:

```js
  request:
   { method: undefined,
     query_string: {},
     headers: {},
     cookies: {},
     data: '<unavailable>',
     url: 'http://<no host>undefined' },
```

image

Ideally, this isn't populated at all since there's no relevant "request" within the context of this worker.

From preliminary digging, it looks like it's because `parsers.parseRequest` adds this info even if it doesn't exist.

LewisJEllis:
Fixes #321.

/cc @MaxBittker

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -238,8 +238,10 @@ extend(Raven.prototype, {
     kwargs.request = extend({}, this._globalContext.request, domainContext.request, kwargs.request);
     if (Object.keys(kwargs.request).length === 0) {
       var req = extend({}, this._globalContext.req, domainContext.req, kwargs.req);
-      var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
-      kwargs = extend(kwargs, parsers.parseRequest(req, parseUser));
+      if (Object.keys(req).length > 0) {
+        var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -238,8 +238,10 @@ extend(Raven.prototype, {
     kwargs.request = extend({}, this._globalContext.request, domainContext.request, kwargs.request);
     if (Object.keys(kwargs.request).length === 0) {
       var req = extend({}, this._globalContext.req, domainContext.req, kwargs.req);
-      var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
-      kwargs = extend(kwargs, parsers.parseRequest(req, parseUser));
+      if (Object.keys(req).length > 0) {
+        var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
+        kwargs = extend(kwargs, parsers.parseRequest(req, parseUser));
```
MaxBittker: do we still want to try to parse user when there's no req?

LewisJEllis: so that user parsing is actually just to look for a user if there's one *on* the req, so it's kind of a nonexistent case; if there's a `user` property in context/kwargs otherwise, we still make use of it.

PR #328, opened 2017-06-05
Adds Node 8 to CI

LewisJEllis:
Gonna go ahead and make sure this passes CI...

### DIFF
--- a/test/raven.parsers.js
+++ b/test/raven.parsers.js
@@ -601,19 +601,17 @@ describe('raven.parsers', function () {
 
     it('should parse an error with additional information', function (done) {
       try {
-        assert.strictEqual(1, 2);
+        var err = new Error('boom');
+        err.prop = 'value';
+        throw err;
       } catch (e) {
         raven.parsers.parseError(e, {}, function (parsed) {
           parsed.should.have.property('exception');
           parsed.exception[0].stacktrace.should.have.property('frames');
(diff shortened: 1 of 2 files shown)


benvinegar:
Err, approved once CI passes, that is.

test/raven.parsers.js
```diff
@@ -601,19 +601,17 @@ describe('raven.parsers', function () {
...
-        assert.strictEqual(1, 2);
+        var err = new Error('boom');
+        err.prop = 'value';
+        throw err;
       } catch (e) {
         raven.parsers.parseError(e, {}, function (parsed) {
           parsed.should.have.property('exception');
           parsed.exception[0].stacktrace.should.have.property('frames');
           parsed.should.have.property('extra');
-          parsed.extra.should.have.property('AssertionError');
```
benvinegar: Should we not be concerned that someone using this assertion library wouldn't work with Raven in Node 8?

LewisJEllis: It'll still work, it just makes the property name `AssertionError [ERR_ASSERTION]` instead because of some new error code stuff (for error "subtypes") in node 8. I didn't want to branch on version so I just fell back to something that checks extra property without "subtype"ing.

benvinegar: Okay. Let's move forward.

Issue #153, opened 2016-05-07
captureException() fails with HTTP Status Code 413


arabold:
I'm browserify & minify my code base into a single huge JS with just a few lines of code (but many, many characters per line obviously). While the code itself works fine, it fails to upload any captured exception to getsentry.com with HTTP Status Code 413.

I could nail it down to the minification. If I'm running the unminified version (still one huge JS file, but this time with many lines of code) then POSTing the exception works as expected. I'm guessing that the reason is that Raven tries to submit callstack information and gets thrown off by the fact that pretty much everything is in a single line of code.

In an ideal world I want to upload my source map to Sentry and have Raven-Node/Sentry figure the rest out. Or maybe I should use `raven-js` instead of `raven-node` although I'm running server-side?

mattrobenolt:
That's definitely the problem. We also don't even support source maps for node. That's exclusively a browser JavaScript feature for us now. 

Since this is a bit confusing to me, can you share some context on why you'd do this for a Node deployment in the first place?

arabold:
https://github.com/arabold/serverless-sentry-plugin -- this is why :)

We're developing projects for AWS Lambda using the popular [Serverless](https://github.com/serverless/serverless) framework. For AWS Lambda it is common to browserify and minify the code base as it improves loading time. The link above is for a plugin that automatically instructs your code to report errors to Sentry.

scottgeye:
Running into the same issue.  Any update or work around for this? 

Thanks!

sid137:
just got bit by this too...   4 hrs of life...

LewisJEllis:
@sid137 sorry to hear that :(

I'll look into this more today, especially since it seems we'll have more folks with long single lines after the recent webpack compatibility workaround in v2.0.0 and the generally increasing popularity of lambda.

Sourcemap support is a relevant topic here, and it's on the roadmap, but it's definitely a bigger undertaking and we don't consider it to be quite *as* crucial of a feature as it is in the browser, for a number of reasons. A fix for the 413 problem specifically should be pretty straightforward, hopefully just a "if line longer than x, truncate" check.

kamilogorek:
Fixed in https://github.com/getsentry/raven-node/commit/8e111ffd1be85ac18e51f446c05ede9f513171cc#diff-50e3aa130a4f97a42ee2cf111c7b1d9d

Also confirmed that stack size for raw express.js app, bundled with browserify and uglifyjs is below 50kb, which fits easily into payload limit.

[...]

PR #329, opened 2017-06-05
Truncate long source context lines

LewisJEllis:
This should address most of the cases causing #153.

/cc @mattrobenolt @MaxBittker

### DIFF
--- a/test/raven.utils.js
+++ b/test/raven.utils.js
@@ -121,6 +121,34 @@ describe('raven.utils', function () {
       }
     });
 
+    it('should trim long source line in surrounding source context', function (done) {
+      var parseStack = raven.utils.parseStack;
+      var callback = function (frames) {
+        var frame = frames.pop();
+        frame.in_app.should.be.true;
+
+        var lineBefore = frame.pre_context[frame.pre_context.length - 1].trim();
+        lineBefore.should.not.startWith('{snip}');
+        lineBefore.should.endWith('{snip}');
+
+        var lineOf = frame.context_line.trim();
+        lineOf.should.startWith('{snip}');
+        lineOf.should.endWith('{snip}');
+        lineOf.length.should.equal(154); // 140 limit + 7 for `{snip} ` and ` {snip}`
+        lineOf.should.containEql("throw new Error('boom');");
+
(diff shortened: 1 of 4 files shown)


test/raven.utils.js
```diff
@@ -121,6 +121,29 @@ describe('raven.utils', function () {
...
+      var callback = function (frames) {
+        var frame = frames.pop();
+
+        frame.pre_context.should.be.an.instanceOf(Array);
+        frame.pre_context[frame.pre_context.length - 1].trim().should.endWith('{snip}');
+        frame.context_line.should.be.type('string');
+        frame.context_line.trim().should.startWith('{snip}');
+        frame.context_line.trim().should.endWith('{snip}');
+        frame.post_context.should.be.an.instanceOf(Array);
+        frame.post_context[0].trim().should.endWith('{snip}');
```
benvinegar: I feel it's not sufficient to just test for presence of `{snip}` – can we improve this test to also assert the content of the snipped code is as expected (e.g. what appears between `{snip}`)?

LewisJEllis:
@benvinegar updated the test. I just check the length and the presence of the `throw` statement; checking exact string equality feels like tautologically testing that the code does what it does. Let me know if there was any other check you had in mind.

benvinegar:
It's an improvement.

> feels like tautologically testing that the code does what it does.

Yes.

dpehrson:
Has this been released yet? Just signed up for sentry (looks awesome) and this was unfortunately the first thing I hit.

We deploy an isomorphic react app that leverages webpack asset url hashing so we webpack the server code as well.

benvinegar:
Not yet. We'll publish something soon.

joevbruno:
@benvinegar any progress / timeline on this?

benvinegar:
@joevbruno – I wanted to do this earlier in the week, but now that it's Friday I'd rather wait until the weekend is over first. There's a bunch of stuff on master that could be problematic, and even this truncation change could cause some issues to group differently. We want to be on-hand just in case.

joevbruno:
@benvinegar Understood. Hopefully it will happen sometime this week

MaxBittker:
this has been released @joevbruno let me know if you have any problems with it!

PR #330, opened 2017-06-09
feat(mergeContext): support deep merge

### LINKED ISSUE #228: mergeContext is shallow and could be better if it went deeper
Right now it just throws our `extend` helper at the context object, which is a shallow non-recursive merge, so:
```javascript
Raven.setContext({
  user: {
    username: 'lewis',
  }
});
Raven.mergeContext({
  user: {
    email: 'lewis@...'
  }
});
console.log(Raven.getContext().user);
```
currently gives us:
```
{
  email: 'lewis@...'
}
```
but we might prefer/expect to get:
```
{
  username: 'lewis',
  email: 'lewis@...'
}
```
It only does mergey stuff in not clobbering any keys on the ctx other than `user`, but it's just replacing the old value of user with the new one.

Right now to accomplish this, we need to do something like:
```javascript
var currUserCtx = Raven.getContext().user;
var newUser = goDoOurOwnMerge(currUserCtx, { email: 'lewis@...' });
Raven.mergeContext({ user: newUser });
```

The context data needs to be json-serializable anyway and we use json-stringify-safe, but we should also be careful not to let merging trip on recursion/circular reference problems. I think just going one level deeper should keep this simple and still make `mergeContext` significantly more helpful, but I haven't really ruled out other options.

/cc @benvinegar @MaxBittker for thoughts

patrickhulce:
fixes #228

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -13,6 +13,7 @@ var domain = require('domain');
 var instrumentor = require('./instrumentation/instrumentor');
 
 var extend = utils.extend;
+var merge = utils.merge;
 
 function Raven() {
   this.breadcrumbs = {
@@ -396,7 +397,7 @@ extend(Raven.prototype, {
   },
 
   mergeContext: function mergeContext(ctx) {
-    extend(this.getContext(), ctx);
+    merge(this.getContext(), ctx);
     return this;
   },
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -45,6 +45,24 @@ module.exports.extend = Object.assign || function (target) {
   return target;
 };
 
+module.exports.merge = function merge(target) {
+  for (var i = 1; i < arguments.length; i++) {
+    var source = arguments[i];
+    for (var key in source) {
+      if (Object.prototype.hasOwnProperty.call(source, key)) {
+        if (target[key] && source[key] &&
+            typeof target[key] === 'object' &&
+            typeof source[key] === 'object') {
+          target[key] = merge(target[key], source[key]);
+        } else {
+          target[key] = source[key];
+        }
+      }
+    }
+  }
+  return target;
+};
+
 module.exports.getAuthHeader = function getAuthHeader(timestamp, apiKey, apiSecret) {
   var header = ['Sentry sentry_version=5'];
   header.push('sentry_timestamp=' + timestamp);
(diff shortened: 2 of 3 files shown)


MaxBittker:
hi @patrickhulce - I appreciate this patch but we're not confident on moving on it right now because of concerns about breaking existing behavior and very deeply nested / circular object tar pits, as well as because of diverging from the raven-js behavior. Might revisit this in the future though

patrickhulce:
alright thanks for the heads up @MaxBittker I implemented it because I needed it, so I'm fine to leave it, lmk if there's any movement on this in the future

arabold:
I'm certainly voting for having this merged in one form or the other in the next major release. Specifically when using `tags` and `extras` it doesn't make sense that `mergeContext` overwrites all previously set values. That's not a merge and actually counterintuitive IMO. A simple fix without processing objects recursively would be to merge `tags` and `extras` individually like this:

```js
function mergeContext(ctx) {
  const oldCtx = this.getContext();
  if (ctx.user) {
    oldCtx.user = ctx.user; // don't merge the user info
  }
  extend(oldCtx.tags, ctx.tags);
  extend(oldCtx.extras, ctx.extras);
  return this;
},
```

I'm happy to provide an updated and tested PR if this would be an option for you.

patrickhulce:
> Specifically when using tags and extras it doesn't make sense that mergeContext overwrites all previously set values.

Agree 100%, I mistakenly believed they would not be overridden when originally using the API. Not sure about the special case handling of no merge for user :), but I'd be happy to update this PR to just shallow extension in @arabold's comment when we're ready.

arabold:
The thought behind not merging the `user` is that you might switch from one user account to another, e.g. if the user logs in and converts from being an anonymous user a specific one. Typically you wouldn't want to merge user properties in those cases.

patrickhulce:
I recognize the `mergeContext({user: null})` workaround isn't as nice for unsetting a user, but it feels a bit odd to have it consistently merge everything except user properties (especially since the motivating example in #228 is for merging user properties). I digress though, perhaps this discussion should move to the issue itself?

kamilogorek:
Closing this for now until we settle on one solution. Let's move discussion to https://github.com/getsentry/raven-node/pull/330 per @patrickhulce comment.

PR #334, opened 2017-06-20
Version 2.1.0

MaxBittker:
@benvinegar thoughts on the patch notes?  existing memory leak tests seemed to pass 👍


benvinegar:
Patch notes look great. Once tests pass, let's 🚢

PR #339, opened 2017-07-12
delete kwargs.req after parsing

ambar:


LewisJEllis:
Good catch, this should definitely be happening; oversight on my part.

benvinegar:
@LewisJEllis – thanks for drawing our attention back to this; totally missed this PR.

benvinegar:
Closing this in favor of #340, which adds a test.

PR #340, opened 2017-07-14
delete kwargs.req after parsing

benvinegar:
See #339


niftylettuce:
Can this please get accepted?  It's been a bug for a long time and your customers are losing important data.

niftylettuce:
Pinging @benvinegar and @MaxBittker - please merge and release a new version asap

LewisJEllis:
> your customers are losing important data.

I don't believe this is causing people to *lose* data, it just causes this processing error warning on the event page for people using the middleware: 


If this *is* in fact causing people to lose data, can you give an example of how?

Issue #346, opened 2017-07-23
_


niftylettuce:

niftylettuce:
What do you guys think of something like this, modifying the `requestHandler` function to accept `isKoa` as a Boolean?

```js
  requestHandler: function (isKoa) {
    var self = this;
    if (isKoa)
      return function(ctx) {
        return new Promise((resolve, reject) {
          var req = ctx.request;
          req.user = ctx.state.user;
          self.context({ req: req }, function () {
            domain.active.add(req);
            domain.active.add(ctx.response);
            resolve();
          });
        });
      }
    return function (req, res, next) {
      self.context({ req: req }, function () {
        domain.active.add(req);
        domain.active.add(res);
        next();
      });
    };
  },
```

niftylettuce:
After we get this in, we need to also update the docs here to show its usage <https://docs.sentry.io/clients/node/integrations/koa/>

benvinegar:
@niftylettuce – I don't like having Koa-specific code inside of the core library (re: `isKoa`). I'd rather we either introduced a Koa-specific middleware, or we just have more elaborate docs on how to make this work.

niftylettuce:
@benvinegar so if we had `requestHandlerKoa` that'd be OK?

benvinegar:
Re-opening – this seems like something we should do.

PR #348, opened 2017-07-23
Added Koa support, Added Mongoose support for user object

niftylettuce:
Don't accept this PR just yet - per #346 and my TODO comment in this pull request we still need to see if we need to `return next()`.  I think we do, but I'm a bit tired at the moment.  Can anyone review this? Thanks.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -49,7 +49,7 @@
     "no-multi-spaces": 2,
     "no-multi-str": 2,
     "no-new": 2,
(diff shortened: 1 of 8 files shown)


niftylettuce:
Travis seemed to randomly fail here <https://travis-ci.org/getsentry/raven-node/jobs/256551916>

niftylettuce:
We need to add tests too for Koa + Mongoose support, I will try to do this.

PR #366, opened 2017-08-30
Update Node flags for v8.X instrumentation tests

kamilogorek:
Added `--expose-internals` flag to all tests and `--expose-http2` for Node >= 8 only, as it's not supported in previous versions.

Included one edge case, where we have to remove http2 flag, as the test itself is asserting it's absence.


benvinegar:
Awesome.

Issue #365, opened 2017-08-30
Dropping support for Node < 4.0


kamilogorek:
Node 0.8 and 0.10 are long time deprecated now, and they require some workarounds for Raven to work, as well for tests to run.

Node 4.0 is still supported for 7 more months, so we should definitely leave it in place for now – https://github.com/nodejs/LTS

Do we have any statistics regarding Node 0.8/0.10 usage @benvinegar? Any thoughts on this?

benvinegar:
It's okay if we move forward and drop support for 0.8 and 0.10. They can always pin to a previous version.

But we should probably add a note to the README commenting that we only support versions according to the LTS schedule.

kamilogorek:
Started initial PR. I'm not sure whether we should wait for major release with this or not.
Do we have any usage stats of Node pre v4?

kamilogorek:
Merged https://github.com/getsentry/raven-node/pull/368

mitjap:
This change should definitely bump major version as it is breaking (maybe just for some users but still).

PR #368, opened 2017-08-31
Pre Node v4 deprecation

kamilogorek:
Initial PR for dropping support of Node pre v4.

- wrote message about supported LTS versions
- removed `<v4` guard from instrumentation tests
- removed v0.10 and v0.12 from TravisCI config

Related issue: https://github.com/getsentry/raven-node/issues/365

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,12 @@
 
 [![Build Status](https://travis-ci.org/getsentry/raven-node.svg?branch=master)](https://travis-ci.org/getsentry/raven-node)
 
+`raven-node` supports only LTS versions of Node, therefore currently required release is `>= v4.0.0`.
+The last known working version for `v0.10` and `v0.12` is `raven-node@2.1.2`.
+Please use this version if you need to support those releases of Node.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,6 +8,12 @@
 
 [![Build Status](https://travis-ci.org/getsentry/raven-node.svg?branch=master)](https://travis-ci.org/getsentry/raven-node)
 
+`raven-node` supports only LTS versions of Node, therefore currently required release is `>= v4.0.0`.
+The last known working version for `v0.10` and `v0.12` is `raven-node@2.1.2`.
+Please use this version if you need to support those releases of Node.
+
+To see up-to-date Node.js LTE Release Schedule, go to https://github.com/nodejs/LTS.
```
HazAT: Typo: LTE Release Schedule -> LTS

kamilogorek:
Fair point @benvinegar. We can most likely get away with not even bumping any version now, as it's not client-facing code.
My only concern would be accidentally using syntax/API not supported in `< 0.12` and still marking it as minor or patch. Then, when someone bumps their dependencies locked to major, can break a build.

PR #369, opened 2017-09-08
Prevent off-heap memory explosion when request is stacking

kamilogorek:
Same code as in https://github.com/getsentry/raven-node/pull/132 + added a simple test


kamilogorek:
Ping

PR #370, opened 2017-09-11
WIP: Add support for sending event requests via proxy

ndmanvar:
Do not merge...just yet.


ndmanvar:
test script:
```
var Raven = require('raven');
Raven.config('https://{PUBLIC_KEY}:{SECRET_KEY}@sentry.io/202440', {
    transport: new Raven.transports.HTTPSProxyTransport({
        proxyHost: 'localhost',
        proxyPort: 8080,
    })
}).install();

var x = {};
x.something();

```

http in URL -> HTTPProxyTransport
https in URL -> HTTPSProxyTransport

lib/transports.js
```diff
@@ -25,6 +27,17 @@ HTTPTransport.prototype.send = function(client, message, headers, eventId, cb) {
     port: client.dsn.port || this.defaultPort,
     ca: client.ca
   };
+
+  // set path apprpriately when using proxy
+  if (this.options.hasOwnProperty('host')) {
```
mfpierre: @ndmanvar I think you're missing setting the `Host` header to be set to `client.dsn.host` in case of proxy.
Without it the host header would keep the proxy host which cause the HTTP request to be rejected by the sentry.io server

ndmanvar:
https://gist.github.com/ndmanvar/7ae7b4db7bf1cda30545c1480fde77fe

ndmanvar:
Will open if revisited/needed.
[Gist](https://gist.github.com/ndmanvar/7ae7b4db7bf1cda30545c1480fde77fe) is the current working solution.

ndmanvar:
Closed this, as I am working off of new branch: https://github.com/getsentry/raven-node/pull/397

Issue #350, opened 2017-07-25
Name of error class retrieved in the wrong way


deefeloper:
Currently, sentry always shows `Error` as the name of the error because in parser,js it says:

`var name = err.name + '';`

https://github.com/getsentry/raven-node/blob/master/lib/parsers.js#L18

This should be changed to:
`var name = err.constructor.name;`

So that when you create custom errors which extend `Error` it will use the correct name

benvinegar:
Makes sense. +1.

PR #372, opened 2017-09-13
fix: Pull Error's name from constructor, not Error itself

kamilogorek:
Fixes: https://github.com/getsentry/raven-node/issues/350

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -15,7 +15,8 @@ module.exports.parseText = function parseText(message, kwargs) {
 
 module.exports.parseError = function parseError(err, kwargs, cb) {
   utils.parseStack(err, function(frames) {
-    var name = err.name + '';
+    var name =
+      ({}.hasOwnProperty.call(err, 'name') ? err.name : err.constructor.name) + '';
     if (typeof kwargs.message === 'undefined') {
       kwargs.message = name + ': ' + (err.message || '<no message>');
     }
--- a/test/raven.parsers.js
+++ b/test/raven.parsers.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var util = require('util');
 var assert = require('assert');
 var raven = require('../');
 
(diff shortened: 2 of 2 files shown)


lib/parsers.js
```diff
@@ -15,7 +15,7 @@ module.exports.parseText = function parseText(message, kwargs) {
 
 module.exports.parseError = function parseError(err, kwargs, cb) {
   utils.parseStack(err, function(frames) {
-    var name = err.name + '';
+    var name = err.constructor.name + '';
```
benvinegar: Should it not be:

```javascript
var name = (err.name || err.constructor.name) + '';
```

That way if someone sets a local name property, that is prioritized over going up the constructor chain?

kamilogorek: If one won't set `err.name` manually, then it defaults to `Error`, which will result in `err.constructor.name` to be never used.

benvinegar: How about:

```javascript
var name = {}.hasOwnProperty.call(err, 'name') ?
  err.name : err.constructor.name;
```

kamilogorek: Got it. Updated and added additional tests covering this cases.

lib/parsers.js
```diff
@@ -15,7 +15,8 @@ module.exports.parseText = function parseText(message, kwargs) {
 
 module.exports.parseError = function parseError(err, kwargs, cb) {
   utils.parseStack(err, function(frames) {
-    var name = err.name + '';
+    var name =
+      ({}.hasOwnProperty.call(err, 'name') ? err.name : err.constructor.name) + '';
```
deefeloper: Please use `Object.prototype.hasOwnProperty.call()` since you don't create a new object first, furthermore I think the constructor name should be leading

kamilogorek: Thanks for pointing that out. However, it's such a micro-optimization, which won't make any difference ever (this object will be collected by GC anyway), that I'll just patch it while working on something else and won't create a separate PR for it. We also use `{}.` notation in other places, and I prefer code consistency over small performance gains :)

Issue #373, opened 2017-09-15
Document overriding server_name with config/ENV


francisdb:
We are running our apps in docker and the server_name that is picked up is just a unique id for the docker container. It would be great if we could customize that field.

This has been available in raven-js for a while by setting `serverName`
https://github.com/getsentry/raven-js/pull/404

kamilogorek:
@francisdb you can already do this, by setting `name` attribute during `config` phase or `SENTRY_NAME` env variable.

https://github.com/getsentry/raven-node/blob/98353fab8f808eaee448f89a90f25a7721434901/lib/client.js#L52
https://github.com/getsentry/raven-node/blob/98353fab8f808eaee448f89a90f25a7721434901/lib/client.js#L247

francisdb:
Docs could be better on this, at least mentions `server_name` like it is visible in the sentry ui. The docs currently mention `SENTRY_NAME` and point to some non-existing python documentation link
https://docs.sentry.io/clients/node/config/

kamilogorek:
Updated the issue. Thanks for pointing that out.

kamilogorek:
Addressed in https://github.com/getsentry/raven-node/pull/377

PR #377, opened 2017-09-22
docs: Add name config description and optional ENV vars

kamilogorek:
Closes https://github.com/getsentry/raven-node/issues/373

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -33,9 +33,21 @@ Those configuration options are documented below:
           logger: 'default'
         }
 
+.. describe:: name
+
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -46,6 +58,7 @@ Those configuration options are documented below:
 .. describe:: environment
 
     Track the environment name inside Sentry.
+    Optionally, use ``SENTRY_ENVIRONMENT`` environment variable.
```
benvinegar: It has been suggested we should default to `NODE_ENV` – super convenient, probably worth exploring.

Issue #353, opened 2017-08-01
Document onFatalError callback available for the install method


simonepri:
We should document this:
https://github.com/getsentry/raven-node/blob/master/lib/client.js#L103

kamilogorek:
Addressed in https://github.com/getsentry/raven-node/pull/378

PR #378, opened 2017-09-22
docs: Include information about fatalError install callback

kamilogorek:
Closes https://github.com/getsentry/raven-node/issues/353

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -6,9 +6,19 @@ To get started, you need to configure Raven to use your Sentry DSN:
 .. sourcecode:: javascript
 
     var Raven = require('raven');
-    Raven.config('___PUBLIC_DSN___').install()
+    Raven.config('___PUBLIC_DSN___').install();
 
 At this point, Raven is ready to capture any uncaught exceptions.
+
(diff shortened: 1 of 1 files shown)


MaxBittker:
FYI, looks like you branched this off preserve-nonenums #379 and it has it's commit

kamilogorek:
Thanks for catching it @MaxBittker, fixed

mitsuhiko (approved):

docs/config.rst
```diff
@@ -6,10 +6,20 @@ To get started, you need to configure Raven to use your Sentry DSN:
...
+``install`` method can optionally take a callback function, in case a fatal error occurs,
+which means that something during capturing initial error went wrong and process is about to exit.
+
+.. sourcecode:: javascript
+
+    Raven.config('___PUBLIC_DSN___').install(function (err, initialErr, eventId) {
+      console.error(err);
+      process.exit(1);
+    });
+
```
benvinegar: Some small language tweaks I'd recommend:

> Note that the ``install`` method can optionally take a callback function that is invoked if a fatal, non-recoverable error occurs.  You can use this callback to perform any cleanup before the Node process exits.
 +which means that something during capturing initial error went wrong and process is about to exit.

Issue #360, opened 2017-08-09
IP address lost when using express behind a proxy and 'trust proxy' is enabled


rwky:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

When `app.enable('trust proxy')` is enabled `req.ip` is `undefined` here https://github.com/getsentry/raven-node/blob/master/lib/parsers.js#L154 this results in `req.connection.remoteAddress` being used.

**What is the expected behavior?**

`req.ip` shouldn't be undefined and used.

I've created a test case:

1) In an empty folder run `npm install express raven`

2) Create index.js with the following:

```js
var http = require('http')
var express = require('express')
var app = express()
app.enable('trust proxy')

var raven = require('raven')

var dsn = '';
raven.config(dsn, {
  autoBreadcrumbs: true
}).install();

app.use(raven.requestHandler());

app.get('/', function(req, res) {
    res.send('hello world ' + req.ip)
})

app.get('/err', function(req, res) { 
    throw new Error('test')
})

app.use(raven.errorHandler());
http.createServer(app).listen(8888);
```

3) Set up a proxy to forward to the node app, I used nginx with the following config:

```
 location / {
                proxy_pass http://172.17.0.1:8888;
                proxy_set_header  X-Forwarded-For $remote_addr;
               proxy_set_header  X-Real-IP       $remote_addr;
        }
```

4) Visit / via the proxy and you'll see hello world and the IP printed

5) Visit /err via the proxy, it'll throw an error, check this in sentry it won't show the IP. I checked that the line mentioned above is resulting in req.ip undefined.

I've not had time to debug why req.ip is undefined but since it's not undefined when accessed outside raven I'm assuming it's a bug in raven.

goorockey:
+1

I found `Object.prototype.hasOwnProperty.call(req, 'ip'))` return false, which leads ip field does not assigned in `extend` and `req.ip` remains undefined in Raven:

[client.js](https://github.com/getsentry/raven-node/blob/master/lib/client.js#L223)
```
      var req = extend({}, this._globalContext.req, domainContext.req, kwargs.req);
      if (Object.keys(req).length > 0) {
        var parseUser = Object.keys(kwargs.user).length === 0 ? this.parseUser : false;
        extend(kwargs, parsers.parseRequest(req, parseUser));
        delete kwargs.req;
      }
```

[utils.js](https://github.com/getsentry/raven-node/blob/master/lib/utils.js#L36)
```
[...]

PR #379, opened 2017-09-22
feat: Preserve some non-enumerable properties from request

kamilogorek:
Fixes https://github.com/getsentry/raven-node/issues/360

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -230,25 +230,13 @@ extend(Raven.prototype, {
       so we only parse a `req` property if the `request` property is absent/empty (and hence we won't clobber)
       parseUser returns a partial kwargs object with a `request` property and possibly a `user` property
     */
-    var initialRequest = utils.prepareInitialRequest(
-      this._globalContext.request,
-      domainContext.request,
-      kwargs.request
-    );
-    kwargs.request = extend(
-      initialRequest,
+    kwargs.request = this._createRequestObject(
       this._globalContext.request,
       domainContext.request,
       kwargs.request
     );
     if (Object.keys(kwargs.request).length === 0) {
-      var initialReq = utils.prepareInitialRequest(
-        this._globalContext.req,
-        domainContext.req,
-        kwargs.req
-      );
-      var req = extend(
-        initialReq,
+      var req = this._createRequestObject(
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -56,6 +56,22 @@ module.exports.getAuthHeader = function getAuthHeader(timestamp, apiKey, apiSecr
   return header.join(', ');
 };
 
+module.exports.prepareInitialRequest = function prepareInitialRequest() {
+  var request = {};
+  var nonEnumberables = ['ip'];
```
kamilogorek: This could be passed as an argument, but I wasn't sure if that's worth it, as it's one-off util.

lib/client.js
```diff
@@ -228,14 +228,29 @@ extend(Raven.prototype, {
       so we only parse a `req` property if the `request` property is absent/empty (and hence we won't clobber)
       parseUser returns a partial kwargs object with a `request` property and possibly a `user` property
     */
+    var initialRequest = utils.prepareInitialRequest(
+      this._globalContext.request,
```
MaxBittker: would you mind adding some comments here? I'm having trouble understanding why it is needed to use both `prepareInitialRequest` and `extend` on these `request` & `req` objects, instead of just `prepareInitialRequest.

test/raven.utils.js
```diff
@@ -98,6 +98,66 @@ describe('raven.utils', function() {
     });
   });
 
+  describe('#prepareInitialRequest', function() {
```
MaxBittker: these tests on the prepareInitialRequest utility seem good to me, but would you mind writing a slightly broader test that recreates the original problem outlined in #360?

MaxBittker (dismissed):
this looks like a good fix, but it's still slightly confusing and I think the comments + test would help.

lib/client.js
kamilogorek: This was just me missing it. Fixed :)

test/raven.utils.js
kamilogorek: Done.

kamilogorek:
@MaxBittker everything has been addressed

mitsuhiko (approved):

Issue #375, opened 2017-09-21
Raven.captureException does not process the request properly in koa app.


chrisjensen:
**What is the current behavior?**

Raven.captureException does not process the request properly in koa app.

Minimum example at [this gist](https://gist.github.com/chrisjensen/75d93f09fb982dc7767ec18d93c88efb).

In the gist, you can see four ways I've tried to capture the request ([lines 29 to 32](https://gist.github.com/chrisjensen/75d93f09fb982dc7767ec18d93c88efb#file-demo-js-L29-L32))

The relavent lines are:

    Raven.captureException(
			err,
			{
    //				req: this.request,
    //				request: this.request,
    //				request: this.req,
				req: this.req,
			},

Test this server by sending a post request to `http://localhost:3000/error` with some body included

Here's the information that's recorded about the request in my sentry dashboard from those different configurations

Request Attachment Method | Body | Headers? | HTTP Method | URL | Other
--- | --- | --- | --- | --- | ---
req: this.req | "\<unavailable\>" | Yes | POST | http://localhost:3000/error | Sentry says: Discarded invalid parameter 'req' 
request: this.req | (none) | Yes | POST | /error | 
request: this.request | (none) | (none) | (none) | (none) | Discarded invalid value for parameter 'request' 
req: this.request | Full Body | (none) | GET (should be POST) | http://\<no host\>/error | Discarded invalid parameter 'req'  

**What is the expected behavior?**

At least one of these ways of attaching the request should cause the full request to be saved, ie:

Request Attachment Method | Body | Headers | HTTP Method | URL | Other
--- | --- | --- | --- | --- | ---
??? | Full Body | Yes | POST | http://localhost:3000/error |

kamilogorek:
Koa itself doesn't provide parsers, you have to attach them yourself, that's why `body` is always empty.

Use https://github.com/koajs/bodyparser and `this.request` to report errors.

Also your implementation is using generators, which will be dropped in v3 of Koa and it's written in a way that it'll never reach `catch` clause (you can verify it using `--inspect` flag and Chrome Devtools).

<img width="1231" alt="screen shot 2017-09-21 at 13 05 49" src="https://user-images.githubusercontent.com/1523305/30692889-99cd0aca-9ecd-11e7-8c74-d1b01b6b6a24.png">

Working example:

```js
'use strict';

const PORT = 3000;

const Koa = require('koa');
const Router = require('koa-router');
const bodyParser = require('koa-bodyparser');

[...]

PR #387, opened 2017-10-05
fix: attach remaining non-enumerables to req

kamilogorek:
Ref: https://github.com/getsentry/raven-node/issues/375

An issue here is that Koa wraps request object, and those properties are not enumerable – https://github.com/koajs/koa/blob/master/docs/api/request.md

`bodyparser` middleware, attaches body to `this.request` object, not `this.req`, therefore we have to use former as the context for the error and we cannot just say `this.request.req` while calling it, eg.

```js
Raven.captureException(err, {
  req: this.request.req,
}, (sendErr, eventId) => {
  if (sendErr) console.error('Failed to send captured exception to Sentry', err);
});
```

This will work, but will skip the `body` of a request, which we definitely don't want.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -526,19 +526,30 @@ extend(Raven.prototype, {
      * (eg. globalContext.request + domainContext.request + kwargs.request),
      * we manually pull them out from original objects.
      *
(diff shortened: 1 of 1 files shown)


MaxBittker (approved):
nice that this was already abstracted 👍

Issue #389, opened 2017-10-06
Integrate with the 'http' library


arasmussen:
There are a bunch of great integrations for those using libraries like Express, etc. We don't use a server framework, just an 'http' server.

It'd be great to have an integration with sentry so that we can easily add context like:
– What is the path of the request?
– What are the query params of the request?
– What is the host header?
– What are the cookies?
– Who is the logged in user?

Of course, I can pass all this in manually, just seems like a lot of work compared to something like:

```es6
http.createServer((request, response) => {
  Raven.setRequestContext(request, (error) => {
    // continue responding to request
  });
});
```

Right now I'm getting errors in Sentry but I'm missing loads of valuable context like which path are they requesting? With what data? Who's the user? (I know I'll have to do the last one on my own anyway). Thanks!

kamilogorek:
You can already achieve it by using wrap/context methods – https://docs.sentry.io/clients/node/usage/#context-wrap

```js
http.createServer(Raven.wrap((request, response) => {
  Raven.setContext({ request: request });
  // continue
}));

or

```js
http.createServer((request, response) => {
  Raven.context(() => {
      Raven.setContext({ request: request });
      // continue
  });
});
```

Most (if not all) data you mentioned should be preserved and send through to Sentry.

arasmussen:
Hey @kamilogorek, thanks for the quick response. But won't that also send a TON of other data that I don't need? That request object seems pretty huge.

It'd also be nice if there were a combined API for wrap + setContext. Separate issue though.

```es6
Raven.context({request: request}, () => {
  // continue
});
```

kamilogorek:
No, it'll be stripped down to only those that we handle on the server :)
https://github.com/getsentry/raven-node/blob/master/lib/parsers.js#L59-L169

arasmussen:
Smart, is that documented somewhere?

arasmussen:
I didn't see it after reading this page https://docs.sentry.io/clients/node/ a few times. If it were in there somewhere, I'd definitely have done that instead of bugging you on this lovely Friday!

kamilogorek:
It indeed was not. Fixed it https://github.com/getsentry/raven-node/pull/390 :)

arasmussen:
Thanks! I just shipped the additional context for our site. Really looking forward to better server error context. Cheers!

kamilogorek:
Cheers! :)

PR #390, opened 2017-10-06
docs: Add more clarification to req/request object

kamilogorek:
ref: https://github.com/getsentry/raven-node/issues/389#issuecomment-334743297

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -136,10 +136,11 @@ All optional attributes are passed as part of the options to ``captureException`
 
     If you're inside a context and your context data includes a ``user`` key, that data will be merged into this.
 
-.. describe:: req
+.. describe:: request
 
-    The ``req`` object associated with this event, from a Node http server, Express, Koa, or similar.
-    Will be parsed for request details and user context from ``req.user`` if present.
+    Alias:  ``req``. The ``request`` object associated with this event, from a Node http server, Express, Koa, or similar.
+    Will be parsed for request details and user context from ``request.user`` if present. It will only pull out the data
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -136,10 +136,12 @@ All optional attributes are passed as part of the options to ``captureException`
...
 
-.. describe:: req
+.. describe:: req/request
 
-    The ``req`` object associated with this event, from a Node http server, Express, Koa, or similar.
-    Will be parsed for request details and user context from ``req.user`` if present.
+    The ``req/request`` object associated with this event, from a Node http server, Express, Koa, or similar.
+    Will be parsed for request details and user context from ``req.user`` if present. It will only pull the data
+    that's handled by the server like method, headers, data, cookies, so you don't need to worry about passing
+    a bloated request object that might contain additional, yet unnecessary data.
```
MaxBittker: >It will only pull the data that's handled by the server like method, headers, data, cookies, so you don't need to worry about passing a bloated request object that might contain additional, yet unnecessary data.

in my opinion, it's better to enumerate the list here and leave it at that, the second part  (not needing to worry) is implied

MaxBittker (approved):
nice improvement, left a suggestion you can take or leave

docs/usage.rst
benvinegar: I'd sooner just name this argument `request` (we accept both, right?), and then mention below that the shorter alias (`req`) is also acceptable. I find labeling this as `req/request` a little confusing.

kamilogorek:
@MaxBittker @benvinegar updated. And yes, we accept both keywords.

PR #397, opened 2017-11-17
Add support for sending events through proxy

ndmanvar:
For both HTTP and HTTPS endpoints. Uses tunnel-agent for https.


lib/transports.js
```diff
@@ -87,12 +96,39 @@ function HTTPSTransport(options) {
...
+}
+
+function HTTPSProxyTransport(options) {
+  this.transport = https;
+  this.options = options || {};
+  this.options.agent = tunnel['httpsOverHttp']({
+    proxy: {
+      host: options.proxyHost,
+      port: options.proxyPort,
+      proxyAuth: null // TODO: Add ability to specify creds/auth
```
ndmanvar: will take care of this (for `HTTPProxyTransport` also)

ndmanvar:
@kamilogorek Can you please review and give me your thoughts/comments? (I know `proxyAuth` is still incomplete/TODO, but wanted to get your thoughts/review on the rest of it)

ndmanvar:
Still need to write tests

ndmanvar:
@kamilogorek can you please review when you get a chance? (:

kamilogorek:
@ndmanvar will do! I just wasn't sure if you're done with the tests :)

test/raven.transports.js
```diff
@@ -15,6 +15,92 @@ describe('transports', function() {
...
+      proxyHost: 'localhost',
+      proxyPort: '8080'
+    };
+
+    // HTTPS
+    var httpsProxyTransport = new transports.HTTPSProxyTransport(option);
+    httpsProxyTransport.options.agent.proxyOptions.should.exist;
+
+    var _cachedAgent = httpsProxyTransport.agent;
+    var requests = {};
```
kamilogorek: You don't seem to use this object anywhere in tests

test/raven.transports.js
```diff
@@ -15,6 +15,92 @@ describe('transports', function() {
     https.agent.maxSockets.should.equal(100);
   });
 
+  it('should set HTTPS proxy transport when proxy config is specified and request is sent', function(done) {
+    var option = {
```
kamilogorek: `s/option/options/g`

lib/transports.js
```diff
@@ -89,10 +122,47 @@ function HTTPSTransport(options) {
   this.options = options || {};
   this.agent = httpsAgent;
 }
+
+function HTTPProxyTransport(options) {
+  this.defaultPort = 80;
+  this.transport = http;
+  this.options = options || {};
+  this.agent = httpAgent;
+  this.options.host = options.proxyHost;
```
kamilogorek: How about just using `host` `port` etc. instead? If we pass `options` object to this constructor, it's by definition mentioned to be used by `proxy`, therefore using a prefix `proxyHost` seems redundant.

kamilogorek: Or it may be useful to distinguish `http` from `httpProxy` options. It's up to you really.

test/raven.transports.js
```diff
@@ -15,6 +15,92 @@ describe('transports', function() {
...
+          host: 'foo',
+          port: 123
+        },
+        emit: function() {}
+      },
+      null,
+      null,
+      null,
+      function() {
+        httpsProxyTransport.options.agent.proxyOptions.headers.host.should.exist;
```
kamilogorek: We could also check that it has been created correctly and assert on `foo:123`

kamilogorek:
@MaxBittker could you also take a peak at this code as well?

lib/transports.js
```diff
@@ -31,14 +34,45 @@ HTTPTransport.prototype.send = function(client, message, headers, eventId, cb) {
...
+      this.options.path =
+        'http://' +
+        client.dsn.host +
+        ':' +
+        client.dsn.port +
+        client.dsn.path +
+        'api/' +
+        client.dsn.project_id +
+        '/store/';
+      delete options.hostname; // only 'host' should be set when using proxy
```
MaxBittker: slightly better comment here for what will happen if you send hostname or a link to a resource that justifies why?

lib/transports.js
```diff
@@ -31,14 +34,45 @@ HTTPTransport.prototype.send = function(client, message, headers, eventId, cb) {
...
+
+      if (this.options.proxyAuth) {
+        // might be able to use httpOverHttp agent
```
MaxBittker: could this be commented to be more explanatory? not sure what these two branches mean

Soliah:
Any updates on this? Would lodging a support request as a paying customer expedite this in anyway?

PR #399, opened 2017-11-27
fix: access 'process' through global variable

kamilogorek:
This fixes correct code behavior when bundled using webpack. Otherwise, our bundled SDK will always yield `undefined` for anything we want to read from `process` variable.

And we do need to access it to gather environment information like `node` version.

Related: https://github.com/getsentry/sentry/pull/6626


kamilogorek:
Umm... lemme fix this test first maybe.

Issue #198, opened 2016-08-27
Explore/document source map support for raven-node/Node projects


benvinegar:
e.g. for users using TypeScript, CoffeeScript, Babel, etc.

benvinegar:
See also getsentry/sentry/issues/2632

benvinegar:
See also: https://github.com/evanw/node-source-map-support

chrisblossom:
Would love to see this added. I am using Babel with node and this would help tremendously trying to track down node-sentry errors. FYI I currently inline my node sourcemap, but do not have to do that. Not sure if that is helpful information.

dpehrson:
Would like to see this as well. We run an isomorphic react app and because the client is webpacked with file-loader for generating asset hashed filenames, the easiest thing to do is just webpack the server before deployment.

Just verified 2.1.0 fixed the issue so that errors can be reported now, but source map support would make them much more useful.

Really neat service so far.

arabold:
I'm using Webpack with the `devtools: source-map` option and load my source maps using `source-map-support` explicitly, i.e. just like this:
```js
"use strict";
require("source-map-support").install();

// ... regular javascript code ...
```

This does indeed work well. For some reason Raven/Sentry isn't able to show the source code itself, but the file names and line numbers are correct. Not sure why this isn't working but due to this I actually switched to `devtools: nosources-source-map` to save a couple of Kbytes in my upload. But the results are the same. It works fine and gives me accurate call stack information.

kamilogorek:
Aaaaand it's done.

PR #400, opened 2017-11-29
docs: Include Source Maps docs

kamilogorek:
This should be enough to get this thing going.

Ref: https://github.com/getsentry/raven-node/issues/198
Related: https://github.com/getsentry/sentry/pull/6626


kamilogorek:
We can merge this only after we actually make Source Maps available :)

bilby91:
@kamilogorek Any ETA for getting this done ? We are using airbrake at the moment but we would happily move to sentry if we get this feature :)

kamilogorek:
@bilby91 it should be ready to go. Would love to know if everything is working for you :)

PR #401, opened 2017-11-29
ref: Enable http instrumentation by default

kamilogorek:
It was omitted by an accident when refactoring a looong time ago.
This change won't break anything, however, not everyone might want to turn this instrumentation on.
On the other hand, it's so minor change, that I don't believe it should wait for the new major release. WDYT?


mitsuhiko (approved):
shyp eat

Issue #396, opened 2017-11-09
Custom breadcrumbs or context is not sent to sentry.


GeeWee:
The code: 

```javascript
Raven.config(
      'https://****@sentry.io/233**',       
     {
         captureUnhandledRejections: true,
     },
);

Raven.context( () => {
    Raven.captureBreadcrumb({
        message: 'Test breadcrumb',
        category: 'payment',
        data: {
            amount: 312,
        }
    });

    Raven.setContext({
        context: 'someContext'
    });
    throw Error(' in context error 2 ')
});
```

Sends the error to sentry, but I'm unable to see the additional context anywhere, and the breadcrumbs aren't available either.

I'm using hosted sentry, and with a raven-node version of "^2.2.1" in my package.json. I'm setting up raven-node as shown above.

kamilogorek:
Hey @GeeWee, there are few things going on here:

- you have to call `install` on `Raven` object in order to use breadcrumbs (after calling `.config`)
- `setContext` is as it's name says, setting a whole context – overriding it, and because breadcrumbs are also part of the error context, they are disappearing
- use `mergeContext` if you want to add some additional values without overriding it
- Sentry supports only specific context keys, `user`, `tags` and `extras` (which are any arbitrary values really) – please refer to https://docs.sentry.io/learn/context/

To summarize, here's a version of your code that will work:

```js
Raven.config(
      'https://****@sentry.io/233**',       
     {
         captureUnhandledRejections: true,
     },
).install();

Raven.context( () => {
    Raven.captureBreadcrumb({
        message: 'Test breadcrumb',
        category: 'payment',
        data: {
            amount: 312,
        }
    });

    Raven.mergeContext({
        extras: {
            moreValues: 'someContext'
        }
    });

    throw Error(' in context error 2 ')
});
```

GeeWee:
You're absolutely right. Thanks a bunch. (Although for anyone else reading, it's `extra` - not `extras`)

I think the behaviour of the following code is a little strange:

```javascript
Raven.config(
  'https://****@sentry.io/**',
		{
			captureUnhandledRejections: true,
		},
	) 
//NOTE NO INSTALL call

Raven.context( () => {
	Raven.captureBreadcrumb({
		message: 'Test breadcrumb',
		category: 'payment',
		data: {
			amount: 312,
		}
	});
	
	Raven.mergeContext({
		extra: {
			moreValues: 'someContext'
		},
		user: {
			foo: 'bar',
		},
[...]

PR #402, opened 2017-11-30
ref: Warn user when using capture/wrap without installing Raven

kamilogorek:
Ref: https://github.com/getsentry/raven-node/issues/396

I don't want to change breadcrumbs behaviour right now, therefore this is the least obtrusive solution.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -49,13 +49,17 @@ extend(Raven.prototype, {
 
     this.raw_dsn = dsn;
     this.dsn = utils.parseDSN(dsn);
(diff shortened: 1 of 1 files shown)


jan-auer (approved):

jan-auer:
flaky test?

MaxBittker:
yup

kamilogorek:
@jan-auer I wouldn't say flaky, as this guard was intentionally put in the code. I'd say it's just lack of any mentions about it in the docs.

Issue #404, opened 2017-12-04
Attach timestamp when capturing message


tmhpfnr:
**Do you want to request a *feature* or report a *bug*?**

I want to request a **feature** / improvement

**Has someone had this problem before?**

Not that I know of.

**What is the current behavior?**

While performing a `raven.captureMessage` always the current timestamp is attached to the underlying `send` method.

**What is the expected behavior?**
It should be possible to pass a timestamp as key argument to adjust when the actual event happened like you already can with breadcrumbs.

kamilogorek:
See: https://github.com/getsentry/raven-node/pull/405#issuecomment-356565953

PR #405, opened 2017-12-04
Feature: Attach timestamp #404

tmhpfnr:


lib/client.js
```diff
@@ -263,7 +263,7 @@ extend(Raven.prototype, {
     kwargs.environment = kwargs.environment || this.environment;
     kwargs.logger = kwargs.logger || this.loggerName;
     kwargs.event_id = eventId;
-    kwargs.timestamp = new Date().toISOString().split('.')[0];
+    kwargs.timestamp = kwargs.timestamp || new Date().toISOString().split('.')[0];
```
kamilogorek: Can you mention this parameter inside `usage` docs in `Additional Data` section? Including desired timestamp format?

test/raven.client.js
```diff
@@ -214,6 +214,30 @@ describe('raven.Client', function() {
...
+          // default behaviour, no timestamp configured
+          client.captureMessage('dafuq is going on');
+        } else {
+          // add timestamp as keyword argument
+          client.captureMessage('dafuq is going on', {
+            timestamp: expectedTimestamp
+          });
+        }
+      });
+      done();
```
kamilogorek: `done` should be moved inside `send` method and triggered only during the second call to make sure that it waits correctly.

kamilogorek: Or can we just streamline this test and get rid of `forEach` at all? We can utilize sinon's `spy/mock` and assert on calls arguments instead. This will make them easier to read and let us remove `done` completely.

kamilogorek:
Thanks for the PR!

kamilogorek:
I'll have to close this one, as I just realized that we need a very specific timestamp format in order for Sentry's system to display information correctly and calculate latencies.

Issue #406, opened 2017-12-06
Discarded invalid value for parameter 'logger' 


valeriangalliat:
I'm getting the following error on all my Sentry events:

> Discarded invalid value for parameter 'logger' 

```
{
  "name": "logger",
  "value": ""
}
```

I never used the `logger` parameter, and it looks like the default value for `logger` as set by `raven-node` itself is `''`: https://github.com/getsentry/raven-node/blob/eb4695a6a7f36320f243611b28bf4597c9de614f/lib/client.js#L71

How to reproduce:

```js
require('raven').Client(process.env.SENTRY_DSN).captureException(new Error('Test'))
```

Outputs:

```js
{ message: 'Error: Tes',
  exception: [ { type: 'Error', value: 'Test', stacktrace: [Object] } ],
  user: {},
  tags: {},
  extra: {},
  breadcrumbs: { values: [] },
  request: {},
  modules: {},
  server_name: 'localhost',
  environment: '',
  logger: '',
  event_id: '...',
  timestamp: '2017-12-06T20:57:31',
  project: '...',
  platform: 'node' }
```

And the "Discarded invalid value for parameter 'logger'" is present on Sentry.

Thank you :)

kamilogorek:
Thanks for the report @valeriangalliat. We made some changes to sentry UI recently. I'll take care of it.

PR #407, opened 2017-12-08
ref: Delete empty properties before sending event to the server

kamilogorek:
Fixes https://github.com/getsentry/raven-node/issues/406

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -54,12 +54,11 @@ extend(Raven.prototype, {
     this.root = options.root || global.process.cwd();
     this.transport = options.transport || transports[this.dsn.protocol];
     this.sendTimeout = options.sendTimeout || 1;
-    this.release = options.release || global.process.env.SENTRY_RELEASE || '';
+    this.release = options.release || global.process.env.SENTRY_RELEASE;
     this.environment =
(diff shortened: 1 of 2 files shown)


jan-auer (approved):

lib/client.js
```diff
@@ -266,11 +265,14 @@ extend(Raven.prototype, {
...
     kwargs.platform = 'node';
+    kwargs.release = this.release;
 
-    // Only include release information if it is set
-    if (this.release) {
-      kwargs.release = this.release;
-    }
+    // Cleanup empty properties before sending them to the server
+    Object.keys(kwargs).forEach(function(key) {
+      if (typeof kwargs[key] === 'undefined' || kwargs[key] === '') {
```
SimonSchick: Why not just use `kwargs[key] === undefined`, the `typeof` check is only really needed for globals.

kamilogorek: I'm just used to use this check instead. It doesn't change any performance characteristics, as it's just one lookup. Although I think we should include `null` here as well, so we should end up with `kwargs[key] == null || kwargs[key] === ''` instead (allowing for `0` to be passed).

PR #408, opened 2017-12-10
feat: Support Boom statusCode

ali-essam:
**What happens?**

raven considers all [Boom](https://github.com/hapijs/boom) errors to be 500 as a Boom error doesn't have statusCode at error root level.

**What should happen?**

raven should try to get the statusCode if error is a Boom error.


kamilogorek:
Wrapped in a parentheses and merged manually. Thanks!

PR #409, opened 2017-12-11
Update link

anteriovieira:


kamilogorek:
Thanks. Agree that this is better one :)

Issue #96, opened 2014-10-09
Express deprecation warning for `req.host`


clayzermk1:
> express deprecated req.host: Use req.hostname instead node_modules/raven/lib/parsers.js:84:19

Messages are showing up in the log for Express 4.9.5.

If I get around to it, I'll cook up a trivial PR and give it some testing love.

Cheers!

Ajar-Ajar:
"express deprecated req.host: Use req.hostname instead ... node_modules/raven/lib/parsers.js:84:19
very easy to fix - I swapped it locally but every time I install the package it throws this error again - please change it. simply replace req.host with req.hostname like the error message suggests.
Thanks
Ajar

oliviertassinari:
This issue resurfaced with this change https://github.com/getsentry/raven-node/commit/5d8556aba2682c8add539a77a8b8f35ae3116797. cc @kamilogorek looks like you recently joined the organization.

kamilogorek:
This is just a warning, not an issue. It'll also never break anything and a change mentioned here https://github.com/getsentry/raven-node/commit/5d8556aba2682c8add539a77a8b8f35ae3116797 is required for some other frameworks.
Is this changing anything for you other than the additional line of text in the logs?

oliviertassinari:
> Is this changing anything for you other than the additional line of text in the logs?

It's not an additional line of text in the logs. It's the first warning we have. It's a noise that will weaken future important warning messages.

yegortokmakov:
guys, please release it

kamilogorek:
@yegortokmakov released as `2.4.0`

PR #412, opened 2017-12-28
Fixed express deprecation warning

AmazingTurtle:
"express deprecated req.host: Use req.hostname instead" has been fixed

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -580,9 +580,9 @@ extend(Raven.prototype, {
     });
     sources = [{}].concat(sources);
     var request = extend.apply(null, sources);
-    var nonEnumberables = [
+    var nonEnumerables = [
       'headers',
-      'host',
+      'hostname',
       'ip',
       'method',
       'protocol',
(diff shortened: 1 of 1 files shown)


oliviertassinari:
Raised into https://github.com/getsentry/raven-node/issues/96#issuecomment-354112329

kamilogorek:
This change will break accessing `host` for Koa users – http://koajs.com/#request so we cannot simply remove this attribute from the table.

oliviertassinari:
[Express is 50 times more used than Koa](https://npm-stat.com/charts.html?package=koa&package=express&from=2016-01-02&to=2017-01-31).

kamilogorek:
@oliviertassinari, unfortunately, it's not enough, but I still wrote a workaround for this. Hope it helps :)
https://github.com/getsentry/raven-node/pull/413
@AmazingTurtle excuse me closing this PR, but it was just faster to write than commenting and asking for a change.

PR #413, opened 2018-01-03
fix: workaround for express.js 'host' deprecation message

kamilogorek:
Ref: https://github.com/getsentry/raven-node/issues/96#issuecomment-354748884
Ref: https://github.com/getsentry/raven-node/pull/412

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -580,9 +580,9 @@ extend(Raven.prototype, {
     });
     sources = [{}].concat(sources);
     var request = extend.apply(null, sources);
-    var nonEnumberables = [
+    var nonEnumerables = [
       'headers',
-      'host',
+      'hostname',
       'ip',
       'method',
       'protocol',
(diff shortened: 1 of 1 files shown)


HazAT (approved):

lib/client.js
```diff
@@ -591,12 +591,24 @@ extend(Raven.prototype, {
...
     });
 
+    /**
+     * Check for 'host' *only* after we checked for 'hostname' first.
+     * This way we can avoid the noise coming from Express deprecation warning
+     * https://github.com/expressjs/express/blob/b97faff6e2aa4d34d79485fe4331cb0eec13ad57/lib/request.js#L450-L452
+     * REF: https://github.com/getsentry/raven-node/issues/96#issuecomment-354748884
+     **/
+    if (!request.hasOwnProperty('hostname')) {
+      sources.forEach(function(source) {
```
AmazingTurtle: Iteration is pointless

AmazingTurtle (changes requested):
This will fix the deprecation warning by express in first place by replacing 'host' by 'hostname' in nonEnumerables. However checking for 'hostname' using hasOwnProperty in the second place does not fix anything at all. Koajs (just to mention, because you made changes for also keeping that one intact) has both hostname and host aswell. Therefore your fix won't apply here.

lib/client.js
kamilogorek: It's not because we want to keep correct order or eventual overrides – https://github.com/getsentry/raven-node/blob/master/lib/client.js#L229-L247

kamilogorek:
@AmazingTurtle in Koa `host` is just `hostname` + `port`. We are mostly concerned about the `hostname` itself, therefore we use similar check here https://github.com/getsentry/raven-node/blob/master/lib/parsers.js#L71-L75

PR #416, opened 2018-01-11
feat: Sensible non-Error exception serializer

kamilogorek:
Initial attempt to tackle  https://github.com/getsentry/raven-node/issues/316 and https://github.com/getsentry/raven-node/issues/187

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -367,8 +367,11 @@ extend(Raven.prototype, {
       if (utils.isPlainObject(err)) {
         // This will allow us to group events based on top-level keys
         // which is much better than creating new group when any key/value change
-        var hash = md5(Object.keys(err).sort());
-        var message = 'Sentry: non-Error exception captured [' + hash + ']';
+        var keys = Object.keys(err).sort();
+        var hash = md5(keys);
+        var message =
+          'Non-Error exception captured with keys: ' +
+          utils.serializeKeysForMessage(keys);
         var serializedException = utils.serializeException(err);
 
         kwargs.message = message;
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -355,20 +356,36 @@ extend(Raven.prototype, {
...
       cb = kwargs;
       kwargs = {};
     } else {
       kwargs = kwargs || {};
     }
 
+    if (!(err instanceof Error)) {
+      if (utils.isPlainObject(err)) {
+        // This will allow us to group events based on top-level keys
+        // which is much better than creating new group when any key/value change
```
benvinegar: Damn, that's a great idea.

lib/client.js
```diff
@@ -355,20 +356,36 @@ extend(Raven.prototype, {
...
+        var hash = md5(Object.keys(err).sort());
+        var message = 'Sentry: non-Error exception captured [' + hash + ']';
```
benvinegar: I don't know that we need to prefix `Sentry:` here.

What about:

```
Non-Error exception captured with keys foo, bar, ...
```

^Something like that, only showing the first 2 or 3 keys.

kamilogorek: Updated

lib/client.js
kamilogorek: image

MaxBittker: ah, that's awesome kamil. do they need to be hashed? I guess it's a length thing, right?

kamilogorek: Without hashing, very large objects will create way too large fingerprints. This way it's always fixed length as you mentioned.

HazAT (approved):

PR #422, opened 2018-01-31
Update LICENSE

dcramer:
- Switch to BSD 3 clause
- Switch byline to Sentry


dcramer:
Maybe we should just up and switch to Apache right now rather than go through this again (since we already need to switch off 2-clause)

dcramer:
i guess i might as well make this 2018

PR #425, opened 2018-02-08
Fix reporting fatal errors

dustinblackman:
## Overview

I have an application in production I noticed was crashing, but never saw any fatal errors sitting in sentry. After digging deeper I found the `uncaughtException` handler doesn't pass a level, so it defaults to `Error`. This PR fixes that.

## Repro

```javascript
const raven = require('raven');

const sentry_dsn = 'INSERT DSN HERE';
const client = raven.Client(sentry_dsn);
client.install();

client.captureException(new Error('This is a regular error'));
throw new Error('This is a fatal error');

console.log('Done'); // Never reached.
```

__Before Fix:__

screen shot 2018-02-08 at 11 28 39 am

__After Fix:__

screen shot 2018-02-08 at 11 19 22 am

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -161,7 +161,7 @@ extend(Raven.prototype, {
         // also we want to make sure we don't go recursion crazy if more errors happen after this one
         firstError = err;
         caughtFirstError = true;
-        self.captureException(err, function(sendErr, eventId) {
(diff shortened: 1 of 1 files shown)


kamilogorek:
Thanks @dustinblackman!

kamilogorek:
Released as `2.4.1`

PR #428, opened 2018-02-17
fix: preserve context in for rejected promises

urish:
currently, context is lost in case of an unhandled promise rejection. This fix keeps track of the context by looking into the rejected promise domain. This fix has been tested in my production environment and seems to do the trick there.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -120,8 +120,9 @@ extend(Raven.prototype, {
 
     if (this.captureUnhandledRejections) {
       var self = this;
-      global.process.on('unhandledRejection', function(reason) {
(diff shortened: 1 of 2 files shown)


kamilogorek:
Thanks! It makes perfect sense. Could you please include a new test that'll verify that context is passed correctly? here's an example of promise test https://github.com/getsentry/raven-node/blob/master/test/raven.client.js#L399-L428

urish:
Okay, I will have a stab at it later

urish:
Here we go - PR updated :)

kamilogorek:
Awesome! Thanks :)

PR #429, opened 2018-02-19
ref: Use named function for middlewares

kamilogorek:
Per https://github.com/getsentry/raven-node/commit/51aa43e6cbc1a57471d55b9f52bf8c33bd64b589#commitcomment-27581611


MaxBittker (approved):
🏷 nice

Issue #430, opened 2018-02-22
Warn/throw when using `captureException` without config


rexxars:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

The following will crash:
```js
const raven = require('raven')
raven.captureException(new Error('eek'))
```

With the following error:
```
/Users/espenh/webdev/myapp/node_modules/raven/lib/client.js:219
    kwargs.user = extend({}, this._globalContext.user, domainContext.user, kwargs.user);
                                                 ^
TypeError: Cannot read property 'user' of undefined
```

**What is the expected behavior?**

It would be great if this instead warned or threw an error with a message telling me that I need to call `config()` before using `captureException`.

rexxars:
Actually not sure if I think it should throw/warn, but I certainly think it shouldn't crash. Created a PR: #431

kamilogorek:
Fixed in https://github.com/getsentry/raven-node/commit/28b1616b5fcd2f2184b82a8d2f27ce870c11ebad

PR #431, opened 2018-02-22
fix: allow calling captureException without configuration

rexxars:
This prevents a crash when using `captureException` without having configured the client first. See #430 for background. A little unsure if this is the wanted behavior, but better than crashing, I figure.


kamilogorek:
It makes perfect sense. Added same test for `captureMessage` and merged it manually. Thanks!

PR #438, opened 2018-03-12
docs: Update README

WanLinLin:
Add the Node version 8.0.0 prerequisite.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -90,6 +90,8 @@ A context might also correspond to, say, a connection lifecycle or a job being h
 Notable keys that you might set include ``user``, ``tags``, and ``extra``.
 These types of extra context data are detailed more under :ref:`raven-node-additional-data`.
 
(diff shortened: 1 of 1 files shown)


kamilogorek:
Hey, I'm happy to add this note, however, I'd need you to ask two things if I may.

a) Move it to more appropriate place - https://github.com/getsentry/raven-node/blob/master/docs/usage.rst#contextwrap at the very bottom of context/wrap section
b) Domains are supported since 0.x. Domains for promises are what's supported since 8.x and they are only used to enhance the context for the current domain. If they're not supported, it'll just be skipped and will use globally set context data. What it means, is that context for regular `context/wrap` and instrumentation is working prior to 8.x just fine. It's just about the promises.

WanLinLin:
Sorry, I didn't express it well. Updated `usage.rst` is committed.

kamilogorek:
Slightly reworded and merged manually. Thanks!

Issue #439, opened 2018-03-20
Do not omit non-standard ports in request url


sameoldmadness:
**Do you want to request a *feature* or report a *bug*?**

report a bug

**Has someone had this problem before?**

no

**What is the current behavior?**

When I am logging a request to with non-standard port, the port is not present in log.

For example, having request to `https://example.org:8080`, I see the following line in the log.

```
https://example.org
```

**What is the expected behavior?**

The url goes to log as is, including the port.

```
https://example.org:8080
```

Maybe [this line](https://github.com/getsentry/raven-node/blob/master/lib/instrumentation/http.js#L25) should be like this:

```js
(options.host || options.hostname || '') // try host before hostname 
```

so if there is a non-standard port in the request url (i.e. `https://example.org:8080`) it would be logged.

sameoldmadness:
Awesome, thanks!

PR #440, opened 2018-03-23
feat: log non-standard ports in breadcrumbs

kamilogorek:
It makes perfect sens for me.

Resolves: https://github.com/getsentry/raven-node/issues/439


billyvg (approved):
Do we need tests that show `:80/:443` get stripped?

kamilogorek:
@billyvg we do have them. `should instrument http to capture breadcrumbs` and `should instrument https to capture breadcrumbs` are covering this cases, as `nock` is using `:80/:443` by default.

PR #445, opened 2018-04-04
fix: Remove a redundant try-catch block

evgenius:
`new Error(message)` creates a stack trace. There's no need to throw and catch it.


kamilogorek:
Nice catch (pun intended)! :D

Tested on:

Node v9
screen shot 2018-04-05 at 14 47 48

Node v4
screen shot 2018-04-05 at 14 48 23

Node v0.12 (just because, we don't support it anyway)
screen shot 2018-04-05 at 14 51 08

PR #452, opened 2018-04-24
fix: shallow-clone context instead of deep-copy

kamilogorek:
Fixes https://github.com/getsentry/raven-node/issues/451

I'm not sure if this should be `2.5.1` or `2.6.0` though. Thoughts?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -333,8 +333,7 @@ extend(Raven.prototype, {
       cb = kwargs;
       kwargs = {};
     } else {
-      // Do not use reference, as we'll modify this object later on
-      kwargs = kwargs ? JSON.parse(stringify(kwargs)) : {};
+      kwargs = utils.isPlainObject(kwargs) ? extend({}, kwargs) : {};
     }
 
     var eventId = this.generateEventId();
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -333,8 +333,7 @@ extend(Raven.prototype, {
       cb = kwargs;
       kwargs = {};
     } else {
-      // Do not use reference, as we'll modify this object later on
-      kwargs = kwargs ? JSON.parse(stringify(kwargs)) : {};
+      kwargs = utils.isPlainObject(kwargs) ? extend({}, kwargs) : {};
```
kamilogorek: `extend({}, 'something')` will treat each letter as a key so we want to avoid it

test/raven.client.js
```diff
@@ -359,7 +340,7 @@ describe('raven.Client', function() {
           foo: null
         }
       };
-      kwargs.extra.foo = kwargs;
+      kwargs.extra.foo = kwargs.extra;
```
kamilogorek: It had to be changed, as referencing `kwargs` directly wouldn't work in this test, as it's not the same reference anymore.

mitsuhiko (approved):
minor release should be fine.

Issue #454, opened 2018-05-02
Write consoleAlerts to stderr


rockymeza:
**Do you want to request a *feature* or report a *bug*?**
feature

**Has someone had this problem before?**
no

**What is the current behavior?**
When Raven is configured with an empty DSN (aka in development), it outputs this to stdout:

    raven@2.4.2 alert: no DSN provided, error reporting disabled

this is very nice and I appreciate it. However when writing scripts that are supposed to write to stdout, this will mix into the output of the script, making it very difficult to use.

**What is the desired behavior?**

Would it be possible to write to stderr for these messages instead of stdout? This could come in the form of just using `console.warn` or some sort of utility to pass in a Console object that was configured differently, like this:

```javascript
const { Console } = require('console');

// somehow pass newConsole to Raven...
// This would not be as easy to use, but I suppose it provides some flexibility.
const newConsole = new Console({ stdout: process.stderr, stderr: process.stderr });
```

I am running raven@2.4.2, but I believe this behavior is still present in master (2.6-ish?).

https://github.com/getsentry/raven-node/blob/381dd6fe69f6c193e8f0fcaf46919efb68dc6e16/lib/utils.js#L126-L137

rockymeza:
Thanks @kamilogorek !

kamilogorek:
@rockymeza released as `2.6.1`

PR #455, opened 2018-05-08
ref: use console.warn for alerts and store them in Set

kamilogorek:
Resolves https://github.com/getsentry/raven-node/issues/454

I totally agree that it should be piped to stderr, and not stdout, as those are always "something went wrong" alerts.

Also Set is supported since Node 4, which is our lowest supported version https://node.green/#ES2015-built-ins-Set.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,7 @@
   "env": {
     "node": true,
     "mocha": true,
+    "es6": true
   },
   "rules": {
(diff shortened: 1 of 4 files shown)


MaxBittker (approved):
This looks good to me - but i noticed that no tests seem to exist for consoleAlertOnce. Maybe worth doing?

PR #456, opened 2018-05-08
ci: Add node 9 and 10 to travis builds

kamilogorek:
I wonder if Travis already supports Node 10 🤔


jan-auer:
Did we always ignore eslint warnings?

kamilogorek:
For `raven-node` yes, we did this since the beginning

Issue #453, opened 2018-05-01
Raven Is Assuming Errors Created From vm is not proper Error and ignore it's stack trace.


kiddkai:
*bug*

**What is the current behavior?**

Currently, we are using [vm](https://nodejs.org/api/vm.html) to create a jail runtime for some of our untrusted scripts. When we run the run the untrusted scripts, they will throw some errors. We capture those errors and try to use:

```js
raven.captureException(e)
```

to sent those errors to sentry. But it seems like it fails the check from this line:

[client.js#369](https://github.com/getsentry/raven-node/blob/master/lib/client.js#L369)

And created a error which a incorrect stack trace. Although they are still legit errors created from a js runtime. 

**What is the expected behavior?**

Because the `Error` in `vm` is not the same `Error` in the main Node runtime, should raven use more generic check rather than `instanceOf Error` for these kind of type checking?

kamilogorek:
Hey @kiddkai, can you provide a reproducible example I could use?

Code below, produces a correct behavior:
```js
const vm = require("vm");
const code = `throw new Error('error inside vm');`;

try {
  vm.runInThisContext(code);
} catch (e) {
  console.log("e.constructor.name:", e.constructor.name);
  console.log("e instanceof Error:", e instanceof Error);
  console.log("e.message:", e.message);
}
```

```
e.constructor.name: Error
e instanceof Error: true
e.message: error inside vm
```

but I'm pretty sure you meant something else, is this correct assuption?

kamilogorek:
Also, the same goes for all the checks inside the vm itself.

```js
const vm = require("vm");
const code = `
(console) => {
  const foo = new Error('error inside vm');
  console.log("foo.constructor.name:", foo.constructor.name);
  console.log("foo instanceof Error:", foo instanceof Error);
  console.log("foo.message:", foo.message);
}
`;

vm.runInThisContext(code)(console);
```

```
foo.constructor.name: Error
foo instanceof Error: true
foo.message: error inside vm
```

kiddkai:
Hey man, sorry for the late reply. We are not using `runInThisContext`, instead, we are using `runInNewContext`.

Here's the example:

https://repl.it/@zekai/random

kiddkai:
Since we are trying to report these kind of errors via `raven.captureException(e)`, it just ignore the original stack trace and gave us a stack trace start from `raven.captureException`, which wasn't what we where looking for :(

kamilogorek:
[...]

PR #457, opened 2018-05-08
fix: correctly detect errors from vm

kamilogorek:
Unify error detection with Raven.js https://github.com/getsentry/raven-js/blob/258c024ff857f48a0d85d814ed1288264d4f7961/src/utils.js#L14-L25
and write appropriate tests for it.

Fixes https://github.com/getsentry/raven-node/issues/453


mitsuhiko (approved):

kiddkai:
👍

PR #458, opened 2018-05-15
ref: Emit transaction instead of culprit

jan-auer:
By default, we emit a `transaction` now instead of the culprit.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -45,10 +45,12 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
       kwargs.extra[name] = extraErrorProps;
     }
 
-    for (var n = frames.length - 1; n >= 0; n--) {
-      if (frames[n].in_app) {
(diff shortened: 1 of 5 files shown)


lib/parsers.js
```diff
@@ -45,10 +45,12 @@ module.exports.parseError = function parseError(err, kwargs, cb) {
...
     }
 
-    for (var n = frames.length - 1; n >= 0; n--) {
-      if (frames[n].in_app) {
-        kwargs.culprit = kwargs.culprit || utils.getCulprit(frames[n]);
-        break;
+    if (!kwargs.transaction && !kwargs.culprit) {
+      for (var n = frames.length - 1; n >= 0; n--) {
+        if (frames[n].in_app) {
+          kwargs.transaction = utils.getTransaction(frames[n]);
```
jan-auer: Changed the logic here to only include a `transaction` if the user hasn't set a `culprit` manually. 
@mitsuhiko  should we add the transaction always - even if a manual culprit is given?

